PR #6, opened 2026-05-08
fix(image): require exact files.slack.com host before attaching auth

Benjamin-Connelly:
## Summary

`teamIDFromFilesURL` decides whether to attach a workspace's `xoxc` Bearer token and `d` cookie to an outbound image fetch. The current check is `strings.Contains(rawURL, "files.slack.com")` plus a path-prefix scan, so a URL whose path or query merely embeds `files.slack.com/files-pri/...` satisfies the gate while still resolving to an attacker-controlled host.

Concretely:

```
https://attacker.com/files.slack.com/files-pri/T01ABCDEF/x.png
https://attacker.com/x?u=https://files.slack.com/files-pri/T01ABCDEF/x.png
https://files.slack.com.attacker.com/files-pri/T01ABCDEF/x.png
https://files.slack.com@attacker.com/files-pri/T01ABCDEF/x.png
```

All four pass the existing check. `authsForURL` then returns the matching team's auth (or, on a foreign team, walks the full fallback list of every configured workspace), and `tryDownload` adds:

```
Authorization: Bearer xoxc-...
Cookie: d=...
```

…to the request to attacker.com. Both halves of the browser-session credential leak in a single fetch.

### Delivery path

`internal/ui/messages/blockkit/image.go` passes the raw block URL straight into the fetcher, so anything that can post a Block Kit `ImageBlock` or attachment `image_url` into a channel the user views can trigger the leak — most realistically a malicious app/bot the workspace has installed, or a hostile peer in a Slack Connect channel.

## Fix

Parse with `net/url` and require `u.Host == "files.slack.com"` before returning a team ID. Path-prefix logic is unchanged. Behavior on canonical URLs is identical; the spoofing shapes now return `""` and the fetch goes out unauthenticated.

## Test plan

- [x] `go test ./internal/image/ -run TestTeamIDFromFilesURL -v` — 13 subcases pass, covering canonical `/files-pri/`, `/files-tmb/`, `/files/` shapes, query-string variants, and the four spoofing vectors above plus empty/garbage/unrelated-host inputs.
- [x] `go test ./...` — full suite passes, no regressions.
- [x] `go vet ./...` — no new warnings.

### DIFF
--- a/internal/image/auth_url_test.go
+++ b/internal/image/auth_url_test.go
@@ -1,0 +1,77 @@
+package image
+
+import "testing"
+
+// TestTeamIDFromFilesURL_HostCheck verifies that team-ID extraction
+// requires the URL host to be exactly files.slack.com. A previous
+// implementation used strings.Contains, which let hostile URLs that
+// merely embedded "files.slack.com/files-pri/..." in their path or
+// query trigger auth attachment, leaking the workspace's xoxc Bearer
+// and 'd' cookie to the attacker.
+func TestTeamIDFromFilesURL_HostCheck(t *testing.T) {
+	cases := []struct {
+		name string
+		url  string
+		want string
+	}{
+		// Legitimate Slack file URLs — must keep working.
+		{
+			name: "files-pri canonical",
+			url:  "https://files.slack.com/files-pri/T01ABCDEF-F0123/foo.png",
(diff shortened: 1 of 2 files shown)


gammons:
thanks!

PR #19, opened 2026-05-19
fix: wrap Kitty image escapes for tmux passthrough

cultlead3r:
Note: This is slop. But it allows images to render native (instead of halfblock) in kitty when tmux is being used. I figured it was worth making a PR for because I assume a lot of people using a terminal slack app are also using tmux with it like me.

## Summary

- wrap Kitty graphics uploads in tmux DCS passthrough when running under tmux
- apply the same wrapping to the Kitty graphics probe
- add tests for tmux wrapping behavior while preserving existing non-tmux output expectations

## Why

When `image_protocol = "kitty"` is forced inside tmux, slk currently emits raw Kitty graphics APC sequences. tmux strips those sequences unless applications wrap them in the `ESC Ptmux; ... ESC \` passthrough envelope with inner ESC bytes doubled. This makes forced Kitty image rendering work inside tmux while leaving `auto` detection behavior unchanged.

## Notes

This intentionally does not change the conservative auto-detection path where tmux still selects halfblock. Users who want native images inside tmux should explicitly set:

```toml
[appearance]
image_protocol = "kitty"
```

## Verification

- `go test ./internal/image ./internal/avatar`
- `go test ./...`
- `go build -o /tmp/opencode/slk-patched ./cmd/slk`
- manually tested in kitty + tmux with `image_protocol = "kitty"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,39 @@ slk --add-workspace
 
 Full walkthrough: [Setup wiki page](https://github.com/gammons/slk/wiki/Setup).
 
+## Inline images in tmux
+
+`slk` can render native Kitty graphics inside tmux, but tmux passthrough
+must be enabled. Add this to your tmux config:
+
+```tmux
+set -g allow-passthrough on 
+```
+
+Then fully restart tmux:
+
+```bash
+tmux kill-server
+```
+
+Inside tmux, verify the active setting:
+
+```bash
+tmux show -gv allow-passthrough
+```
+
+Expected output is `all` (or `on`).
+
+To force Kitty image rendering in `slk`, set:
+
+```toml
+[appearance]
+image_protocol = "kitty"
+```
+
+`image_protocol = "auto"` remains conservative inside tmux and falls back
+to half-block rendering.
+
 ## Debugging
 
 Set `SLK_DEBUG=1` to enable a comprehensive debug log written to
--- a/README.md
+++ b/README.md
@@ -72,36 +72,26 @@ Full walkthrough: [Setup wiki page](https://github.com/gammons/slk/wiki/Setup).
 
 ## Inline images in tmux
 
-`slk` can render native Kitty graphics inside tmux, but tmux passthrough
-must be enabled. Add this to your tmux config:
+If you run `slk` inside tmux on a Kitty-capable terminal (Kitty, Ghostty,
+WezTerm), images render natively as long as tmux passthrough is enabled:
 
 ```tmux
-set -g allow-passthrough on 
+set -g allow-passthrough on
 ```
 
-Then fully restart tmux:
-
-```bash
-tmux kill-server
-```
-
-Inside tmux, verify the active setting:
+Reload tmux for the setting to take effect (`tmux kill-server`, then
+reattach). Verify with:
 
 ```bash
 tmux show -gv allow-passthrough
 ```
 
-Expected output is `all` (or `on`).
+Expected output: `on` (or `all`).
 
-To force Kitty image rendering in `slk`, set:
-
-```toml
-[appearance]
-image_protocol = "kitty"
-```
-
-`image_protocol = "auto"` remains conservative inside tmux and falls back
-to half-block rendering.
+If passthrough is off, `slk` detects this at startup and falls back to
+half-block rendering automatically — no config change needed. To force a
+specific renderer regardless of detection, set `image_protocol` in
+`config.toml` to `kitty`, `sixel`, `halfblock`, or `off`.
 
 ## Debugging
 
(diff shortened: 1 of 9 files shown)


gammons:
hi @cultlead3r thanks for the PR! 

I tested this locally and I couldn't get it to work.  Here's what I noticed running kitty + tmux:

1. running `slk` without setting the config.toml appearance setting shows the halfblock images.
2. running `slk` with setting the config.toml appearance setting **still** shows the halfblock images, and the slk app got really sluggish for me.  I got this same experience with `ghostty` as well.

I was testing with tmux v3.6a.   Also for yucks I deleted my tmux conf, but the behavior I was seeing did not change.

cultlead3r:
Forgot to mention that I love the app.

When you say setting the appearance in config.toml, you explicitly set it to "kitty"? This is mine:

```toml
[appearance]
image_protocol = "kitty"
```

Also, do you have tmux passthrough enabled? This is also required. On my machine in tmux I have:

```bash
set -g allow-passthrough all
```

I think this might be the reason it didn't work for you. The default in tmux is off.

For me, the working setup is:
- kitty
- tmux 3.6a
- image_protocol = "kitty"
- allow-passthrough = all

The halfblocks and sluggish behavior you describe is exactly what I experienced too before this patch.

gammons:
Perfect, this works.  2 things:

1. Is there a downside to making this the default behavior? (e.g. no `image_protocol` config)
2. It's worth adding a thing to README.md about the tmux config needed.

cultlead3r:
1. The only downside that comes to mind is if someone is using kitty + tmux but they don't have passthrough enabled, they're going to get that sluggish behavior. RIght now, if they leave it on the default auto, it's either going to just work (kitty by itself) or it's going to render in halfblock (kitty+tmux) -- but they're not going to encounter that random perplexing sluggishness unless they explicitly forced kitty as the image protocol.

Probably just adding something in the README warning about tmux is fine -- I guess you can assume that if they're modifying the config file and found "kitty" as an image_protocol config option, they will see the tmux config note as well.

2. I added a section to the README.

gammons:
I went ahead and tested the "make it just work" path end-to-end before pushing back too hard. Sharing what I found — turned out to be more nuanced than I expected.

**The detection problem.** I removed the `env.TMUX != ""` short-circuit in `Detect()` expecting auto-detect to fall through to the kitty branch. It didn't, because tmux is more aggressive about stripping terminal identity than I assumed:

```
TMUX="/tmp/tmux-1001/default,..."
TERM="tmux-256color"      ← overwritten
TERM_PROGRAM="tmux"       ← overwritten
KITTY_WINDOW_ID=""        ← didn't propagate
```

None of the three signals `Detect()` uses for kitty (`KittyWindowID`, `TERM=xterm-kitty`, `TERM_PROGRAM in {ghostty,WezTerm}`) are visible inside tmux. The auto path always returned halfblock and the probe never ran. Forcing `image_protocol = "kitty"` worked because it bypassed env entirely — which is exactly why you were defaulting to the config-driven approach.

**The fix.** Ask tmux. `tmux display-message -p '#{client_termname}'` returns the attached client's actual TERM (`xterm-kitty` in my setup). When it contains kitty/ghostty/wezterm we propose `ProtoKitty`, then your probe confirms whether passthrough is actually on. If the probe times out → halfblock fallback, no behavior change for users without passthrough.

End-to-end result on my box (kitty + tmux 3.6a + `allow-passthrough on`, no config change):

```
[imgrender] tmux client_termname="xterm-kitty"
[imgrender] image protocol detect: cfg="auto" result=kitty
[imgrender] image protocol: kitty
```

Images render natively. Total detection cost ~8ms.

**On the sluggishness.** It's gone once detection works correctly. My read: it was the placement cells referencing image IDs the terminal never received (because tmux was stripping the unwrapped escapes). With detection + your wrapping both in place, the app is snappy.

**On the README.** Reframed your section to lead with `allow-passthrough on` as the prereq rather than the `image_protocol = "kitty"` config knob, since auto-detect now handles the common case. The config option still works as a manual override and I called that out at the end.

**Two follow-up commits** are pushed to `gammons/slk:fix/tmux-kitty-images` for you to cherry-pick if you'd like to include them in this PR:

```
git fetch https://github.com/gammons/slk.git fix/tmux-kitty-images
git cherry-pick FETCH_HEAD~1 FETCH_HEAD
```

That picks `cd62962` (the tmux detect change) and `08da7df` (the README rewrite). Diff: https://github.com/gammons/slk/compare/main...fix/tmux-kitty-images

If you'd rather I take it from here and merge a separate follow-up PR, that's fine too. Just let me know how you'd like to land it. Thanks again — the escape-wrapping is the core of this fix and I wouldn't have caught the env-stripping issue without your testing.

cultlead3r:
Okay sweet. I grabbed your two commits and pulled them in. Should be ready to merge now.

gammons:
nice work @cultlead3r !

Issue #5, opened 2026-05-06
Enterprise App Token Support


focusaurus:
I'm on enterprise slack at work and following the regular token instructions got me booted out of all my login sessions, sent notices to my admins, etc. Presumably this is a security/compliance policy we have in place. Do you know if slk would need to be made available in the official slack app marketplace such that it could be potentially approved for use in a given enterprise workspace?

gammons:
hmm, that's quite a thing!  I am wondering how Slack Enterprise knows that a non-official client is in use.  Regarding an official slack app - are you able to install Slack apps?  if so, then that might be a viable option.

focusaurus:
Unsure if I personally can install apps into our instance but probably not. But we do have many apps installed.

tebriel:
https://github.com/rneatherway/gh-slack fetches the enterprise token from the app bundle's cookie jar using a decryption key fetched from the keychain on MacOS, maybe there's some secret sauce there?

gammons:
I don't have a way a verifying, since I have no Enterprise slack accounts. Maybe @focusaurus if you could take a look at `gh-slack` and uh, see if you get yelled at by your IT team 😄

raff:
I also tried with an Enterprise account. It kind of worked (i.e. I was able to login in slk, but I didn't see any message - though it's possible that I picked a workspace with no messages). Then, when I closed 'slk' it logged me out of Slack (all instances) and got an email (from Slack) "[Security notice] You've been signed out of Slack":

Hello,

We recently detected potentially suspicious activity associated with your Slack account. As a precaution, we’ve signed you out of the following Enterprise Grid organization:
   
   User: XXXXX
   Organization: YYYYY

tebriel:
The repo I linked works with enterprise slack but idk if maybe only for a subset of apis. Who knows how they’re detecting that stuff 🤷🏻

antoszka:
> I'm on enterprise slack at work and following the regular token instructions got me booted out of all my login sessions, sent notices to my admins, etc. Presumably this is a security/compliance policy we have in place. Do you know if slk would need to be made available in the official slack app marketplace such that it could be potentially approved for use in a given enterprise workspace?

I can confirm I have the same issue, `slk` got stuck on `Connecting`, and I got booted out of the web sessions and received the same e-mail as quoted by @raff.

gammons:
[...]

PR #22, opened 2026-05-20
Add browser-like headers to all Slack-bound traffic (refs #5)

gammons:
## Summary

Addresses #5. Decorates every outbound request to `*.slack.com` with the
HTTP headers a recent desktop Chrome on `app.slack.com` would send, so
slk's `xoxc`-token traffic looks indistinguishable from the official
browser client at the header level. Goal: stop Enterprise Grid anomaly
detectors from signing users out / triggering security alerts.

## What changed

- **New `internal/slackhttp` package** — `BrowserTransport`
  (`http.RoundTripper`) that adds `User-Agent` (per-OS Chrome 120),
  `Accept`, `Accept-Language`, `Origin`, `Referer`, and `Sec-Fetch-*`
  headers to requests bound for `*.slack.com` hosts. Never overrides
  caller-set headers (so the per-team `Authorization: Bearer xoxc-...`
  + `Cookie: d=...` on the image fetcher and the hand-rolled endpoints
  survive). Non-Slack hosts pass through untouched.
- **Slack-go HTTP client** (`internal/slack/client.go`) now constructs
  its inner `*http.Client` via `slackhttp.NewBrowserHTTPClient`, so
  every Web API call AND every hand-rolled endpoint POST inherits the
  headers for free.
- **WebSocket upgrade** carries the same headers via a new
  `wsUpgradeHeaders()` helper, with `Sec-Fetch-Dest: websocket` to match
  what a real browser sends on a WS open. (Gorilla doesn't go through
  `http.RoundTripper`, so the headers are passed inline to `Dialer.Dial`.)
- **Image fetcher** drops the `slk/inline-image-fetcher` User-Agent (a
  giveaway just as flagable as `Go-http-client/1.1`) and now goes
  through `BrowserTransport` via the client constructed in `main.go`.
- **README** has a new Enterprise Grid section with honest framing:
  best-effort mitigation, please file an issue if you still get booted.

## Out of scope (intentionally)

- **TLS fingerprint matching (uTLS).** Re-assess only if header parity
  doesn't move the needle.
- **OAuth / Slack App route.** Would cripple core features (no access
  to undocumented endpoints like `client.counts`,
  `users.channelSections.list`, etc.) and require enterprise admin
  approval. Tracked separately.
- **User-configurable UA override.** YAGNI; add only if a real user
  needs it.

## Test plan

Automated (all passing in CI / locally):

- [x] `go test ./... -race -count=1` — all 39 packages pass
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] Unit tests for `BrowserTransport` cover positive (Slack-host
      requests get all 8 headers), negative (non-Slack hosts untouched),
      subdomain matrix (`slack.com`, `files.slack.com`,
      `hackclub.enterprise.slack.com`, `wss-primary.slack.com`),
      caller-header preservation (UA / Authorization / Cookie), nil
      Header/URL defense, per-GOOS UA selection
- [x] Integration tests assert the wiring contracts: slack-go's client
      transport is `*slackhttp.BrowserTransport`; WS upgrade carries
      `User-Agent: Mozilla/...`, `Origin: https://app.slack.com`,
      `Accept-Language`, and `Sec-Fetch-Dest: websocket`; image fetcher
      no longer emits the legacy `slk/inline-image-fetcher` UA
- [x] No stray references to `slk/inline-image-fetcher` or
      `Go-http-client` in production code

Manual (maintainer eyes required before broad announce — see
`docs/superpowers/plans/2026-05-20-browser-like-headers-verification-notes.md`):

- [ ] Run slk against a non-Enterprise workspace with mitmproxy in
      front; visually confirm every Slack request carries the expected
      headers and that neither `Go-http-client` nor
      `slk/inline-image-fetcher` appears
- [ ] Functional smoke test: onboarding → channels → send → receive →
      image attachment → thread reply
- [ ] After merge: post the draft #5 comment asking
      @focusaurus / @raff / @antoszka to retest on Enterprise Grid

## Important caveats

This may not be sufficient. The next escalation is TLS fingerprinting
(Go's ClientHello looks nothing like Chrome's), which would require
uTLS. We won't know whether header parity is enough until an Enterprise
Grid user retests. The README and the draft #5 comment frame this
honestly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,25 @@ slk --add-workspace
 
 Full walkthrough: [Setup wiki page](https://github.com/gammons/slk/wiki/Setup).
 
+## Enterprise Grid
+
+slk authenticates via the same `xoxc` browser session token and `d`
+cookie that `app.slack.com` uses. Every outbound request to `*.slack.com`
+is decorated with browser-like headers (`User-Agent`, `Origin`,
+`Referer`, `Sec-Fetch-*`) so Slack's anomaly detectors should treat the
+traffic the same way they treat the browser tab the token was extracted
+from.
+
+This is best-effort mitigation, not a contract. Your IT policy may still
+flag slk regardless. If you're on Enterprise Grid and slk signs you out
+or triggers a security email after login, please file an issue with:
+
+1. The exact email or notification text Slack sent.
+2. Whether you got logged out of *all* sessions or only slk's.
+3. The output of `slk --version`.
+
+See [#5](https://github.com/gammons/slk/issues/5) for history.
+
 ## Inline images in tmux
 
 If you run `slk` inside tmux on a Kitty-capable terminal (Kitty, Ghostty,
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -8,7 +8,6 @@ import (
 	"image"
 	"io"
 	"log"
-	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -27,6 +26,7 @@ import (
 	"github.com/gammons/slk/internal/notify"
 	"github.com/gammons/slk/internal/service"
 	slackclient "github.com/gammons/slk/internal/slack"
+	"github.com/gammons/slk/internal/slackhttp"
 	"github.com/gammons/slk/internal/ui"
 	"github.com/gammons/slk/internal/ui/channelfinder"
 	"github.com/gammons/slk/internal/ui/compose"
@@ -550,7 +550,8 @@ func run() error {
 		})
 		log.Printf("image fetcher: registered team %q (%s) for file auth", t.TeamName, t.TeamID)
 	}
-	imageHTTPClient := &http.Client{Timeout: 10 * time.Second}
+	imageHTTPClient := slackhttp.NewBrowserHTTPClient(nil)
+	imageHTTPClient.Timeout = 10 * time.Second
 	imageFetcher := imgpkg.NewFetcher(imageCache, imageHTTPClient)
 	imageFetcher.SetAuths(auths)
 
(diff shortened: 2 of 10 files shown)


gammons:
**Update — fix pushed in response to user retest feedback**

@focusaurus / @raff / @antoszka: an Enterprise Grid user retested the
initial commit and reported partial success — login and channel-list
worked, but selecting a channel logged them out of all clients. That's
useful: login and channel-list go through \`slack-go\`, which puts the
\`xoxc\` token in the **form body** (\`token=xoxc-...\`). Channel-select
fires several hand-rolled endpoints in rapid succession, and all of
them were using \`Authorization: Bearer xoxc-...\` instead.

Bearer is an OAuth pattern. Real browsers never send Bearer to
\`app.slack.com\`. Once we started sending browser-shaped headers
(\`Origin: https://app.slack.com\`, Chrome UA, \`Sec-Fetch-*\`), the
combination of "browser headers + Bearer" became a contradictory
request signature — strictly more anomalous than the previous "Go UA +
Bearer" shape, because that one was at least internally consistent.

Commit [7aaf302](https://github.com/gammons/slk/pull/22/commits/7aaf302)
moves the token to the form body and removes the Authorization header
for all five hand-rolled endpoints (\`client.counts\`,
\`conversations.mark\`, \`subscriptions.thread.mark\`,
\`users.prefs.get\`, \`users.channelSections.list\`,
\`subscriptions.thread.getView\`). The image fetcher's Bearer for
\`files.slack.com\` is a different surface (file CDN) and is unchanged
for now — channel-select doesn't trigger image loads.

Empirical evidence the new shape works on Enterprise: it's exactly what
\`slack-go\` already does, and the same user reported \`slack-go\`-driven
calls (login, channel-list) were fine.

If you're up for another round of testing on Enterprise Grid, this fix
needs your hands — there's no Enterprise workspace we can verify against.

Issue #13, opened 2026-05-19
Feature request: Save thread to markdown file


jftuga:
## Problem

One thing I struggle with in the native Slack app is cleanly copying an entire thread and saving it to a text file with decent formatting. Slack's copy/paste produces a mess of inline timestamps, avatar placeholders, and broken whitespace that requires manual cleanup every time.

## Proposed solution

A keybinding (`S`) that, when focused on a thread panel, saves the full thread (parent message + all replies) to a CommonMark markdown file on disk.

The output would look like:

```markdown
**Alice** — 2026-05-18 3:04 PM
Hey, did anyone look at the deploy failure?

**Bob** — 2026-05-18 3:05 PM
Yeah, it was the config change from yesterday.
I reverted it and redeployed.

**Alice** — 2026-05-18 3:06 PM
Thanks! Confirmed it's green now.
```

Key details:

- Slack mrkdwn converted to CommonMark (bold, strikethrough, links, code blocks)
- @mentions and #channels resolved to display names
- Attachments rendered as markdown links
- Reactions rendered as unicode emoji with counts
- Files saved to `~/.local/share/slk/exports/` (respects `XDG_DATA_HOME`)
- Filename pattern: `slk-thread-{channel}-{YYYY-MM-DD-HHMMSS}.md`
- Status bar toast confirms the save

## Current status

I've vibe-coded a working PR for this with Claude Opus 4.6, but since the changeset is somewhat large (new `internal/export` package, a Slack-mrkdwn-to-CommonMark converter, keybinding wiring, and tests) I wanted to check if there's interest before submitting it.

Happy to share the PR if this is something you'd want in slk.

gammons:
@jftuga I think it's an interesting feature.  One of the goals for this project is to create a slack environment that techies would like.  I think techies would like this feature, so it would be a good addition.

jftuga:
PR https://github.com/gammons/slk/pull/40 has merged this feature.

Issue #2, opened 2026-05-01
Enterprise workspaces don't work


taciturnaxolotl:
Hi! Love the project! When using with enterprise Slack workspaces, the api endpoints are different, causing the client to never fully connect. API routes have enterprise before the workspace name, e.g. `https://hackclub.enterprise.slack.com/api/conversations.view`

`slk 0.6.2 (commit a7615cb10bd72959ae5d08c7ad670da6cb5f6f6f, built 2026-05-01T01:08:07Z)`

gammons:
I just cut `v0.6.3` which should address this issue. Let me know.

taciturnaxolotl:
works! thank you!

t3kkm0tt:
@taciturnaxolotl well, I tried logging into hackclub slack and it doesn't work.

taciturnaxolotl:
hmm i haven't tried the newer versions but it did work for me on the hc slack in `0.6.3`

Issue #1, opened 2026-04-30
Can't switch between workspaces


msmiel:
First, thanks for making this. 

I can't seem to get the workspace switching to work. Both the number and workspace selector don't work in either of the terminals I've tried.

<img width="324" height="570" alt="Image" src="https://github.com/user-attachments/assets/e6d00eb2-18f6-470a-83dc-5d4fbc62caa3" />

--- 
slk v 0.6.1   commit 53c9a222

Terminals:
Ghostty v 1.3.1  build 15212  commit 332b2aefc

Rio v 0.4.0  commit 77b1c9c

MacOS 26.4.1 (25E253)

gammons:
Hi @msmiel what happens when you hit `ctrl+w`?  do you see both workspaces?

msmiel:
I see both workspaces, can enter the other workspace's name, but the sidebar and my chat doesn't update if I hit enter.

<img width="531" height="481" alt="Image" src="https://github.com/user-attachments/assets/2792d4af-6196-4d96-9408-bee3fb71acda" />

gammons:
try removing `~/.local/share/slk` and see if that fixes things.

msmiel:
Just wanted to confirm that fixed it. Thanks!

PR #40, opened 2026-05-27
feat: save thread to markdown file with S keybinding

jftuga:
# feat: save thread to markdown file with `S` keybinding

> **Disclaimer:** This PR was 100% vibe-coded by Claude Opus 4.6.

Issue: https://github.com/gammons/slk/issues/13

~415 lines across 10 files (2 new, 8 modified).

## Summary

- Adds a new `S` keybinding (normal mode, thread panel focused) that saves the entire thread (parent + all replies) to a CommonMark markdown file
- Files are saved to `~/.local/share/slk/exports/` (or `$XDG_DATA_HOME/slk/exports/`) with the naming pattern `slk-thread-{channel}-{YYYY-MM-DD-HHMMSS}.md`
- Slack mrkdwn is converted to CommonMark: `*bold*` becomes `**bold**`, `~strike~` becomes `~~strike~~`, `<url|label>` becomes `[label](url)`, mentions and channels are resolved to display names
- Attachments are rendered as markdown links and reactions as unicode emoji with counts
- Status bar shows a confirmation toast with the filename on save

## New files

- `internal/export/markdown.go` — Thread-to-markdown assembler and XDG export directory helper
- `internal/export/markdown_test.go` — Tests for thread export formatting

## Modified files

- `internal/ui/messages/render.go` — Added `SlackMrkdwnToCommonMark()` converter (sibling to `RenderSlackMarkdown`, same input format, CommonMark output instead of ANSI)
- `internal/ui/messages/render_test.go` — 12 tests for the CommonMark converter
- `internal/ui/thread/model.go` — Added `UserNames()` and `ChannelNames()` accessors
- `internal/ui/statusbar/model.go` — Added `ThreadSavedMsg` and `ThreadSaveFailedMsg` types
- `internal/ui/keys.go` — Added `SaveThread` binding (`S` key)
- `internal/ui/mode_normal.go` — Key dispatch for `SaveThread`
- `internal/ui/reducer_io.go` — Toast handling for `ThreadSavedMsg` and `ThreadSaveFailedMsg`
- `internal/ui/app.go` — `saveThreadToFile()` method and `sanitizeForFilename()` helper

## Test plan

- [X] `go build ./...` compiles cleanly
- [X] `go test ./internal/ui/messages/ ./internal/export/` passes all new and existing tests
- [X] Open slk, navigate to a thread, press `S` — status bar shows "Saved slk-thread-{channel}-{date}.md"
- [X] Verify file exists in `~/.local/share/slk/exports/`
- [X] Open the saved file — confirm bold, links, mentions, code blocks, reactions, and attachments render correctly as CommonMark
- [X] Press `S` when not focused on the thread panel — confirm no action taken
- [X] Press `S` on an empty thread panel — confirm no action taken

### DIFF
--- a/internal/export/markdown.go
+++ b/internal/export/markdown.go
@@ -1,0 +1,80 @@
+// Package export converts slk message data into portable file formats.
+// It has no UI dependencies (no lipgloss, no bubbletea) and operates
+// entirely on the messages.MessageItem data model.
+package export
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	emojiutil "github.com/gammons/slk/internal/emoji"
+	"github.com/gammons/slk/internal/ui/messages"
+	emoji "github.com/kyokomi/emoji/v2"
+)
+
+// ThreadToMarkdown converts a parent message and its replies into a
+// CommonMark markdown string. userNames and channelNames resolve
+// @mentions and #channel references in message bodies.
+func ThreadToMarkdown(parent messages.MessageItem, replies []messages.MessageItem, userNames, channelNames map[string]string) string {
+	var b strings.Builder
+	b.WriteString(formatMessage(parent, userNames, channelNames))
+	for _, r := range replies {
+		b.WriteString(formatMessage(r, userNames, channelNames))
+	}
+	return strings.TrimRight(b.String(), "\n") + "\n"
+}
+
+func formatMessage(msg messages.MessageItem, userNames, channelNames map[string]string) string {
+	var b strings.Builder
+
+	b.WriteString("**" + msg.UserName + "**")
+	b.WriteString(" — ")
+	b.WriteString(msg.DateStr + " " + msg.Timestamp)
+	b.WriteByte('\n')
+
+	body := messages.SlackMrkdwnToCommonMark(messages.MessageTextSource(msg), userNames, channelNames)
+	b.WriteString(body)
+	b.WriteByte('\n')
+
+	if len(msg.Attachments) > 0 {
+		for _, att := range msg.Attachments {
+			label := att.Name
+			if label == "" {
+				if att.Kind == "image" {
+					label = "Image"
+				} else {
+					label = "File"
+				}
+			}
+			b.WriteString("[" + label + "](" + att.URL + ")\n")
+		}
+	}
+
+	if len(msg.Reactions) > 0 {
+		parts := make([]string, 0, len(msg.Reactions))
+		for _, r := range msg.Reactions {
+			e := emoji.Sprint(":" + emojiutil.StripSkinTone(r.Emoji) + ":")
+			parts = append(parts, fmt.Sprintf("%s %d", strings.TrimSpace(e), r.Count))
+		}
+		b.WriteString(strings.Join(parts, "  "))
+		b.WriteByte('\n')
+	}
+
+	b.WriteByte('\n')
+	return b.String()
+}
+
+// ExportDir returns the default directory for saved exports,
+// honoring XDG_DATA_HOME. Creates nothing — callers must MkdirAll.
+func ExportDir() (string, error) {
+	if dir := os.Getenv("XDG_DATA_HOME"); dir != "" {
+		return filepath.Join(dir, "slk", "exports"), nil
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".local", "share", "slk", "exports"), nil
+}
--- a/internal/export/markdown_test.go
+++ b/internal/export/markdown_test.go
@@ -1,0 +1,68 @@
+package export
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/gammons/slk/internal/ui/messages"
+)
+
+func TestThreadToMarkdown_ParentOnly(t *testing.T) {
+	parent := messages.MessageItem{UserName: "alice", DateStr: "2026-05-18", Timestamp: "3:04 PM", Text: "hello world"}
+	got := ThreadToMarkdown(parent, nil, nil, nil)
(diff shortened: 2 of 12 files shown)


jftuga:
I know this is a big and somewhat intrusive PR so totally understand if it is too unwieldy to merge and thus maintain from now on. I need this feature and thought I would share it per [your suggestion](https://github.com/gammons/slk/issues/13#issuecomment-4493416110). Please let me know if it is "close" and I can adjust it accordingly.

gammons:
Thanks for the contribution, @jftuga — really nice scope on this one. The `internal/export` package is cleanly layered (no UI deps, just operates on `MessageItem`), and reusing the existing regex catalog in a sibling `SlackMrkdwnToCommonMark` keeps both renderers in lockstep going forward. Tests cover the happy paths well, the build/vet are clean, and the keybinding + reducer wiring follows the established patterns.

I'd like to see two things addressed before merge:

### 1. Emoji shortcodes are resolved inside inline code & code fences

In `internal/ui/messages/render.go` (end of `SlackMrkdwnToCommonMark`):

```go
output = emoji.Sprint(emojiutil.StripSkinToneFromText(output))
```

This runs *after* the inline-code / code-block placeholders are restored, so shortcodes inside backticks get converted to unicode. Neither Slack nor CommonMark does this. I confirmed with a quick probe:

```
input:  "see `:smile:` here"
output: "see `😄 ` here"        // expected: "see `:smile:` here"

input:  "```\n:smile:\n```"
output: "```\n😄 \n```"          // expected: verbatim
```

`TestCommonMark_InlineCodePreserved` only verifies `*foo*` survives, so it doesn't catch this. The fix is to run `emoji.Sprint`/`StripSkinToneFromText` inside `slackMrkdwnToCommonMarkInline` (per-line, while placeholders are still active) rather than on the post-restore output. Worth adding the two cases above as regression tests.

Minor side-note: `emoji.Sprint(":smile:")` returns `"😄 "` with a trailing space (a kyokomi quirk). Invisible in flowing text but jumps out inside backticks. A `strings.TrimRight(..., " ")` on the resolved token cleans it up.

### 2. Non-ASCII channel names produce empty or `--` filenames

`sanitizeForFilename` in `internal/ui/app.go` only keeps `[A-Za-z0-9_]`, which drops every character of a non-ASCII channel name:

| input    | sanitized | filename                  |
|----------|-----------|---------------------------|
| `general` | `general` | `slk-thread-general-…md` ✓ |
| `café`    | `caf`     | `slk-thread-caf-…md` (lossy) |
| `日本語`   | `""`      | `slk-thread--…md` (double dash) |

Two small changes fix this:
- Use `unicode.IsLetter(r) || unicode.IsDigit(r)` instead of the ASCII ranges so Japanese / German / Cyrillic / etc. channel names keep their identity.
- When the sanitized name ends up empty, fall back to a sentinel (or skip the dash entirely) so you never produce the `slk-thread--…md` shape.

`sanitizeForFilename` is a pure helper — easy to add a small table-driven test alongside the fix.

---

A few smaller things I noticed that are fine as follow-ups (not blocking):

- Filename uses second-precision (`2006-01-02-150405`); two rapid `S` presses silently overwrite. Bumping to ms or using `O_EXCL` would close this.
- `wiki/Keybindings.md` doesn't list `S` yet — would be good to add a row.
- Pressing `S` outside the thread panel is a silent no-op; a short toast ("Open a thread first") would be more discoverable.
- The "Saved …" toast shows only `filepath.Base(path)`, so users can't see *where* the file went without knowing XDG conventions.

Thanks again — looking forward to the next iteration!

jftuga:
I am working on #1 and #2, but will need help testing.

I want to push back on: Filename uses second-precision
* The user has to navigate to a thread, press `S`, then navigate to a different thread and press `S` again, all within one second. That's not realistic in a TUI.

My idea for: The "Saved …" toast shows only filepath.Base(path), so users can't see where the file went without knowing XDG conventions.
* `Saved ~/.local/share/slk/exports/slk-thread-general-2026-05-27-143052.md` is 72 chars in length. On a typical 120-col terminal that leaves maybe 50-60 chars for the toast. It'll fit but be tight and could get clipped on narrower terminals. On 80 cols it would definitely get cut off or displace the help hint.
* What if we just show `Saved to ~/.local/share/slk/exports/` (the directory) in the toast, dropping the filename. The user can see which file it is by `ls -t` on that directory. The directory path is the more useful unknown here.  Do you agree with this idea?

gammons:
@jftuga some thoughts:

1. totally fine on the second-precision thing.  I agree with your notion there
2. given the majority of users of `slk` should be comfortable in the terminal, I don't think we need to take any action here.  we can just have the docs say where it's going to be saved.

jftuga:
@gammons Can you please test #1 and #2?

### 1. Emoji shortcodes inside code spans/fences (fixed)

Moved emoji resolution from the post-restore output (where inline code and code block placeholders had already been swapped back in) into `slackMrkdwnToCommonMarkInline`, which runs while placeholders are still active. Also trims the trailing space that `kyokomi/emoji` appends after each resolved glyph. Added three regression tests covering inline code, code blocks, and trailing space.

### 2. Non-ASCII channel names in filenames (fixed)

`sanitizeForFilename` now uses `unicode.IsLetter`/`unicode.IsDigit` instead of ASCII ranges, so channels like `café`, `日本語`, etc. preserve their names. Empty results (all-special-character names) fall back to `"unknown"`. Added a table-driven test.

### Non-blocking items

- **`wiki/Keybindings.md`**: Added `S` row with the export directory path documented.
- **Silent no-op on `S` outside thread**: Now shows an "Open a thread first" toast.
- **Second-precision filenames**: Agreed this is fine as-is.
- **Full path in toast**: Per your suggestion, no action taken — directory is documented in the wiki instead.

jftuga:
@gammons 

I also ran into merge conflicts with your latest version, but these are now resolved (via Claude).

gammons:
Thank you @jftuga looks great!

PR #59, opened 2026-06-08
feat: add keybindings to resize sidebar width

jftuga:
## Summary

> This feature was 100% vibe coded with Claude Opus 4.6.

- Add `]` / `[` keybindings (normal mode) to widen/narrow the sidebar panel
- Width adjusts in 5-column increments, clamped to a 20–60 column range (default remains 30)
- The help screen (`?`) automatically reflects the new bindings

## Motivation

Long channel names get truncated at the hardcoded 30-column sidebar width. There was no configuration, keybinding, or runtime option to adjust it. This adds a quick vim-friendly way to resize without leaving the keyboard.

## Changes

- `internal/ui/sidebar/model.go` — add `width` field, `GrowWidth()`/`ShrinkWidth()` methods, and width constants
- `internal/ui/keys.go` — add `SidebarGrow` (`]`) and `SidebarShrink` (`[`) to `KeyMap`
- `internal/ui/mode_normal.go` — wire the new keybindings to the sidebar model

## Known pre-existing issue

The down arrow key does not scroll past the first page of entries in the help overlay (`?`). This is **not** introduced by this PR — `internal/ui/mode_help.go` and `internal/ui/help/model.go` are untouched on this branch (last modified in commit `933ae44`). The bug likely lives in how `tea.KeyDown` is mapped or consumed in the help modal's key handler.

## Test plan

- [X] Run `go build ./...` — passes
- [X] Run `go test ./internal/ui/sidebar/ ./internal/ui/` — all 360 tests pass
- [X] Launch slk, press `]` repeatedly — sidebar widens in 5-col steps up to 60
- [X] Press `[` repeatedly — sidebar narrows down to 20
- [X] Verify long channel names are no longer truncated at wider widths

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -825,6 +825,34 @@ func run() error {
 		}
 	})
 
+	// Wire sidebar width saver: always persist to the active workspace.
+	app.SetWidthSaver(func(width int) {
+		if activeTeamID == "" {
+			return
+		}
+		teamName := activeTeamID
+		if wctx, ok := workspaces[activeTeamID]; ok && wctx.TeamName != "" {
+			teamName = wctx.TeamName
+		}
+		tomlKey := activeTeamID
+		for k, w := range cfg.Workspaces {
+			if w.TeamID == activeTeamID {
+				tomlKey = k
+				break
+			}
+		}
+		if cfg.Workspaces == nil {
+			cfg.Workspaces = make(map[string]config.Workspace)
+		}
+		ws := cfg.Workspaces[tomlKey]
+		ws.TeamID = activeTeamID
+		ws.SidebarWidth = width
+		cfg.Workspaces[tomlKey] = ws
+		if err := saveWorkspaceWidth(configPath, tomlKey, activeTeamID, teamName, width); err != nil {
+			log.Printf("save workspace sidebar width: %v", err)
+		}
+	})
+
 	// Wire presence/DND status setter. Captured workspaces map and
 	// activeTeamID by reference so the closure always targets the
 	// currently-active workspace context.
@@ -1380,6 +1408,7 @@ func run() error {
 			TeamID:           wctx.TeamID,
 			TeamName:         wctx.TeamName,
 			Theme:            cfg.ResolveTheme(teamID),
+			SidebarWidth:     cfg.ResolveWidth(teamID),
 			Channels:         wctx.Channels,
 			FinderItems:      wctx.FinderItems,
 			UserNames:        wctx.UserNames,
@@ -1532,6 +1561,7 @@ func run() error {
 				TeamID:           wctx.TeamID,
 				TeamName:         wctx.TeamName,
 				Theme:            cfg.ResolveTheme(wctx.TeamID),
+				SidebarWidth:     cfg.ResolveWidth(wctx.TeamID),
 				Channels:         wctx.Channels,
 				FinderItems:      wctx.FinderItems,
 				UserNames:        wctx.UserNames,
--- a/cmd/slk/save_theme.go
+++ b/cmd/slk/save_theme.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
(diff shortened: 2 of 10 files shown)


gammons:
This works great and feels natural — nice addition. One thing I'd love to see (and would be happy for you to drive): **persisting the width across restarts.** Right now `GrowWidth`/`ShrinkWidth` only mutate in-memory state, so the panel resets to 30 on every launch.

The cleanest path is to follow the **existing theme persistence pattern** exactly, which already does global-default + per-workspace-override:

1. **Storage** — add a global default `width` under `[sidebar]` in `config.toml` (the `Sidebar` struct already exists in `internal/config/config.go`), and a per-workspace override `width` on the `Workspace` struct (alongside the existing `theme` field).

2. **Resolution** — add a `ResolveWidth(teamID)` mirroring the existing `ResolveTheme()` in `internal/config/config.go` (per-workspace value wins, else global default, else `defaultWidth`). Load the result into `sidebar.Model.width` at startup.

3. **Save** — add `saveGlobalWidth` / `saveWorkspaceWidth` mirroring `saveGlobalTheme` / `saveWorkspaceTheme` in `cmd/slk/save_theme.go` (the surgical textual-edit approach that preserves comments/order), wired via a callback from the `[`/`]` resize path like the theme saver is wired in `cmd/slk/main.go`.

**One design decision worth calling out:** themes have an explicit scope prompt at selection time ("this workspace" vs "global"), but `[`/`]` are incremental nudges with no such moment. The simplest faithful analog is to have `[`/`]` always persist to the **active workspace** (`saveWorkspaceWidth`), with `[sidebar] width` as the read-time fallback default — no scope prompt needed.

Minor: each keypress would trigger a full `config.toml` rewrite (same as the theme saver does today, so it's a proven-fine pattern). A small debounce is optional polish, not a blocker.

jftuga:
Implemented width persistence following the theme persistence pattern exactly as described.  See the second and third comments.

## What changed

- `ResolveWidth(teamID)` mirrors `ResolveTheme()`: per-workspace `sidebar_width` > global `[sidebar] width` > default 30
- `saveWorkspaceWidth` / `saveGlobalWidth` in new `save_width.go` — same textual-rewrite approach as `save_theme.go`
- `[`/`]` always persist to the active workspace (no scope prompt)
- Width is loaded from config on workspace ready and workspace switch via the existing message pipeline

One naming deviation from the original suggestion: used `sidebar_width` (not `width`) on the `Workspace` struct/TOML key to avoid ambiguity with other widths in the codebase. Under `[sidebar]`, it's just `width` since the section already scopes it.

## Test plan

- [X] `go build ./...` passes
- [X] `go test ./internal/ui/sidebar/ ./internal/ui/ ./internal/config/ ./cmd/slk/` — all tests pass
- [X] Launch slk, press `]` several times, quit, relaunch — sidebar width is restored
- [X] Press `[` to narrow, quit, relaunch — narrower width is restored
- [X] Inspect `config.toml` after resize — `sidebar_width = N` appears under the active workspace block
- [X] Add `width = 40` under `[sidebar]` in config.toml, remove any per-workspace `sidebar_width`, relaunch- sidebar starts at 40
- [X] With both global `[sidebar] width` and per-workspace `sidebar_width` set, verify per-workspace wins
- [X] With neither set, verify sidebar defaults to 30
- [X] Switch workspaces (if multi-workspace) — each workspace restores its own saved width

## NOT TESTED

@gammons Do you have a way of testing this:

- [ ] Resize in workspace A, switch to workspace B (no saved width), verify B uses global default (not A's width)

## Bug found during testing

`saveGlobalTheme` and `saveWorkspaceTheme` in `save_theme.go` silently failed when `config.toml` did not exist — `os.ReadFile` returned an error and both functions bailed without creating the file. This meant that if a user deleted their config (or ran slk for the first time without one), theme changes were never persisted. Fixed both functions to treat a missing file as empty content and create the parent directory if needed. The same fix was applied preemptively to `save_width.go`.

gammons:
Works perfectly.  thanks @jftuga !

PR #64, opened 2026-06-08
Reduce binary size via compiler flags

jftuga:
## Summary

- Add `-ldflags="-s -w"` to strip the symbol table and DWARF debug info from the binary
- Add `-trimpath` to remove local filesystem paths from the compiled binary, improving build reproducibility
- update 'go install' instructions in `README.md`
- update 'go build' in `.github/workflows/ci.yml`

## Test plan

- [X] Run `make build` and verify it completes without errors
- [X] Compare the binary size before and after (expect a meaningful reduction)
- [X] Confirm `bin/slk` runs correctly
- [X] Confirm `go install -ldflags="-s -w" -trimpath github.com/gammons/slk/cmd/slk@latest` runs correctly 
- [X] Confirm `go build -ldflags="-s -w" -trimpath ./...` runs correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
           cache: true
 
       - name: Build
-        run: go build ./...
+        run: go build -ldflags="-s -w" -trimpath ./...
 
       - name: Test
         run: go test ./... -race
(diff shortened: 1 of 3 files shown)


gammons:
Love it!

PR #66, opened 2026-06-10
Fix reaction invalid_name: source emoji from iamcal/emoji-data, drop kyokomi

### LINKED ISSUE #65: Emoji reactions silently fail with invalid_name (picker uses non-Slack shortcodes)
## Summary

Adding an emoji reaction from the picker often fails silently. The reaction appears in your own client (optimistic update), but **other users never see it**, and it **disappears when you switch channels and come back**.

## Steps to reproduce

1. Select a message and press `r` to open the reaction picker.
2. Pick a common emoji such as 👍, 👌, or ✅.
3. The reaction shows locally, but other clients never receive it; navigate away and back and it's gone.

## Root cause

The reaction picker is built from `kyokomi/emoji`'s `CodeMap()` (~4772 shortcodes), which carries CLDR-style aliases like `thumbs_up`, `OK_hand`, `check_mark`. Slack's `reactions.add` does **not** accept those and returns `invalid_name`; it only accepts its own canonical short_names (`+1`, `ok_hand`, `heavy_check_mark`, …).

Two things hid the failure:

- **The API error was swallowed** — `ReactionSentMsg` was a no-op, so a failed `reactions.add` had no surface (`internal/ui/reducer_reactions.go`).
- **The reaction cache is only written by the `reaction_added` WebSocket echo** (`OnReactionAdded` → `db.UpsertReaction`). A failed API call → no echo → nothing persisted, so the optimistic-only reaction vanishes on reload.

Confirmed with `SLK_DEBUG=1`:

```
reaction: firing emoji="thumbs_up" remove=false ...
Warning: reaction API call failed: invalid_name
```

## Why kyokomi is the wrong source

Slack's standard emoji are defined by [`iamcal/emoji-data`](https://github.com/iamcal/emoji-data) (maintained by a Slack co-founder; it's the dataset Slack's own client bundles), keyed by Unicode codepoint with one canonical `short_name` per glyph. There is **no Slack API** for standard emoji names — `emoji.list`, `cache/.../emojis/info`, and `cache/.../emojis/search` are all custom-emoji-only (verified: `emojis/info` returns `failed_ids: ["+1","joy",...]`). kyokomi's superset of gemoji+CLDR aliases includes names Slack never accepts.

Verified against the real web client: Slack sends `name=+1` for 👍 and gets `{"ok":true}`.

## Fix

Source standard-emoji data from `iamcal/emoji-data` (a bundled, generated table) instead of `kyokomi/emoji`, mirroring how Slack's own client works:

- The picker offers only Slack-valid names, so `reactions.add` can no longer produce `invalid_name`.
- Workspace custom emoji shadow standard names and are sent verbatim (e.g. a custom `:facepalm:` must not be rewritten to the standard `:face_palm:`).
- Surface API failures: roll back the optimistic update and log, instead of leaving a phantom reaction.
- Glyphs derived from iamcal codepoints are byte-for-byte identical to kyokomi's for rendering, so message/sidebar/export output is unchanged; the `kyokomi/emoji` dependency is dropped entirely.

Frodotus:
Closes #65.

## Problem

Adding an emoji reaction from the picker silently failed for many emoji: it showed locally (optimistic update) but other users never saw it and it vanished on channel switch. Root cause: the picker was built from `kyokomi/emoji`'s `CodeMap()`, whose CLDR-style aliases (`thumbs_up`, `OK_hand`, `check_mark`) are rejected by Slack's `reactions.add` as `invalid_name`. The error was swallowed (`ReactionSentMsg` was a no-op), and the reaction cache is only written by the `reaction_added` WebSocket echo — so a failed call left nothing behind.

## Fix

Source standard-emoji data from `iamcal/emoji-data` (the dataset Slack's own client bundles; there is no Slack API for standard-emoji names — `emoji.list`/`emojis.info`/`emojis.search` are custom-only), via a generated table, and drop `kyokomi/emoji` entirely:

- **Picker offers only Slack-valid names**, so `reactions.add` can no longer produce `invalid_name`. Slack's own client sends `+1` for 👍 — now so do we.
- **Custom emoji shadow standard names** and are sent verbatim (a custom `:facepalm:` is not rewritten to standard `:face_palm:`).
- **API failures are surfaced**: the optimistic update is rolled back and logged instead of leaving a phantom reaction.
- **`CodeMap()`/`Sprint()` are faithful drop-ins** — glyphs derived from iamcal codepoints are byte-for-byte identical to kyokomi's (verified across simple, VS16, keycap, ZWJ-family, and flag emoji), so message/sidebar/export rendering is unchanged.
- The generated table is committed and refreshable via `go generate ./internal/emoji/...` (provenance: `iamcal/emoji-data`).

## Behavior change

The picker now uses Slack's exact (iamcal) names, so a few emoji shed kyokomi's gemoji-style aliases — e.g. `rainbow_flag` → `rainbow-flag`, `man_astronaut` → `male-astronaut`. This matches what Slack emits and accepts; incoming rendering is unaffected (iamcal's names are a superset of everything Slack sends), and ~2800 junk CLDR/skin-tone variants drop out of the picker.

## Testing

- `go test ./...` passes; added coverage for canonicalization (incl. custom-emoji shadowing) and CodeMap glyph parity.
- Verified end-to-end against a real workspace in a kitty terminal: standard + custom reactions apply, are visible to others, persist across channel switches, and render correctly inline and in reaction pills.


gammons:
Nice fix — root-cause analysis is clear and the iamcal table + glyph-parity approach looks right. A few things worth tightening before merge:

**1. Rollback is wired into only 1 of 4 reaction paths.** The new `ReactionSentMsg` rollback fields are only populated in the picker path (`mode_reaction_picker.go`). The three toggle paths still emit bare `ReactionSentMsg{Err: err}`:
- `toggleReactionOnSelectedMessage` (`app.go:773,778`)
- `toggleReactionOnSelectedThread` (`app.go:801,806`)
- `toggleReactionOnMessageItem` / click-on-pill (`app.go:830,835`)

On failure the reducer runs `updateReactionOnMessage("", "", "", "", true)` — empty IDs match nothing, so the phantom reaction persists and nothing is logged. These toggle already-canonical names so `invalid_name` is unlikely, but network/rate-limit failures still reproduce the original bug. Worth making consistent (or documenting why only the picker needs it).

**2. A few comments now contradict the code:**
- `reactionpicker/model.go:290` — "the picker is built from kyokomi/emoji's CLDR-aliased CodeMap" sits right below the line changed to `slkemoji.CodeMap()`.
- `slacknames.go:12` — `CanonicalSlackName` doc still says the picker is populated from kyokomi.
- `reducer_reactions.go` header block still says `ReactionSentMsg ... No-op today ... a failed call has no surface`, contradicting the rollback added below it.

**3. No test for the headline behavior** — the rollback-on-failure path in `reduceReactions` has no coverage. Canonicalization and glyph parity are tested, but the actual fix (surface failure + roll back) isn't.

Minor: the generator fetches `iamcal/emoji-data` from `master` (`gen_slacknames.go:36`), so `go generate` isn't reproducible — consider pinning to a tag/SHA.

Verified locally: build clean, `go vet` clean, full suite + `-race` on emoji/reaction packages all pass.

Frodotus:
Thanks for the thorough review! Addressed all four in `1f98d17`:

1. **Rollback in all reaction paths** — wired the `ReactionSentMsg` rollback fields into the three toggle paths (`toggleReactionOnSelectedMessage` / `toggleReactionOnSelectedThread` / `toggleReactionOnMessageItem`); they were emitting bare `ReactionSentMsg{Err: err}` and rolling back with empty IDs.
2. **Stale comments** — fixed all three (the `reducer_reactions.go` header `ReactionSentMsg` description, `reactionpicker/model.go:290`, and the `CanonicalSlackName` doc) to match the iamcal sourcing + rollback behavior.
3. **Test for the headline behavior** — added `internal/ui/reducer_reactions_test.go` covering both rollback-on-failure and keep-on-success.
4. **Reproducible generator** — pinned `gen_slacknames.go` to `iamcal/emoji-data` `v16.0.0`; regenerating produces a byte-identical table, so `go generate` is now reproducible.

Build, `go vet`, and the full suite are green.

gammons:
Thanks, nice work!

PR #70, opened 2026-06-10
Show online/away presence for DM peers in the sidebar

### LINKED ISSUE #69: Direct Messages list doesn't show who's online (peers' presence never subscribed)
## Summary

The Direct Messages list never shows who's online — every DM renders the hollow "away" dot (`○`), even for users who are currently active.

## Root cause

Most of the presence pipeline already exists — the WS `presence_change` handler (`OnPresenceChange` → `PresenceChangeMsg` → `sidebar.UpdatePresenceByUser`), the `ChannelItem.Presence` field, and the sidebar rendering (`●` when `Presence=="active"`, else `○`). Two pieces were missing:

1. **slk only subscribes to its own presence.** `bootstrapPresenceAndDND` calls `SubscribePresence([]string{wctx.UserID})` (self only) and never subscribes to DM peers, so Slack never sends `presence_change` for them. Every DM stays at its initialized `"away"`.

2. **Batch `presence_change` events were dropped.** The connection is opened with `batch_presence_aware=1`, so Slack sends `presence_change` with a **`users` array**, but `wsPresenceEvent` only parsed a single `user` field — so even a manual subscription wouldn't have updated anything.

## Fix

- Subscribe to presence for self **plus every 1:1 DM peer** in a single `presence_sub` (it replaces the prior set, so self and peers must go together). Wired into initial boot (once DM peers are known) and `OnConnect` (re-subscribes per connection on reconnect).
- Parse both the single-user (`user`) and batched (`users[]`) forms of `presence_change`.
- `presence_sub` pushes each subscribed user's current presence back, so the subscription doubles as the initial fetch — no per-user query needed.

## Result

DM rows show a filled `●` for online peers and a hollow `○` for away, updating live as people come and go.

Frodotus:
Closes #69.

## Problem

The Direct Messages list never showed who was online — every DM rendered the hollow `○` (away), even for active users.

## Root cause

Most of the presence pipeline already existed (WS `OnPresenceChange` → `PresenceChangeMsg` → `sidebar.UpdatePresenceByUser`, the `ChannelItem.Presence` field, and `●`/`○` rendering). Two pieces were missing:

1. **slk only subscribed to its own presence** (`SubscribePresence([]string{wctx.UserID})`), never DM peers — so Slack never sent their `presence_change` events.
2. **Batch `presence_change` events were dropped.** With `batch_presence_aware=1`, Slack sends `presence_change` carrying a `users` array, but the parser only read a single `user` field.

## Fix

- Add `subscribeWorkspacePresence`: subscribes self + every 1:1 DM peer in a single `presence_sub` (it replaces the prior subscription set, so self and peers must be sent together). Wired into `connectWorkspace` once the DM peers are known (initial boot) and `OnConnect` (re-subscribes per connection, required because the subscription is connection-scoped). App/bot and group DMs are skipped (no human presence dot).
- Parse both the single-user (`user`) and batched (`users[]`) forms of `presence_change`.
- `presence_sub` pushes each subscribed user's current presence back, so the subscription doubles as the initial fetch — no per-user query needed.

## Testing

- `go test ./...` passes; added `TestWorkspacePresenceIDs` (self + DM peers, dedup, skips channels/group/app DMs).
- Verified end-to-end in a kitty terminal: DM rows show `●` for online peers and `○` for away, updating live.

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -95,10 +95,10 @@ func (a sectionsProviderAdapter) OrderedSlackSections() []sidebar.SectionMeta {
 
 // WorkspaceContext holds all state for a single connected workspace.
 type WorkspaceContext struct {
-	Client      *slackclient.Client
-	ConnMgr     *slackclient.ConnectionManager
-	RTMHandler  *rtmEventHandler
-	UserNames   map[string]string
+	Client     *slackclient.Client
+	ConnMgr    *slackclient.ConnectionManager
+	RTMHandler *rtmEventHandler
+	UserNames  map[string]string
 	// AvatarURLs maps userID -> avatar image URL. Populated from the
 	// local users cache at connect time (synchronous, before any
 	// goroutines spin up) and refreshed from the background
@@ -122,7 +122,7 @@ type WorkspaceContext struct {
 	// background users.list fetch and any on-demand resolveUser calls.
 	// Used during channel construction to bucket app DMs into a separate
 	// "Apps" sidebar section.
-	BotUserIDs        map[string]bool
+	BotUserIDs map[string]bool
 	// SectionStore holds the user's Slack-native sidebar sections for
 	// this workspace. Nil when use_slack_sections is disabled, the
 	// REST bootstrap failed, or this workspace hasn't connected yet.
@@ -151,7 +151,7 @@ type WorkspaceContext struct {
 	// after a failed one. The UI uses it to decide whether to draw
 	// the "Threads list unavailable" banner.
 	SubscriptionsAvailable bool
-	Channels    []sidebar.ChannelItem
+	Channels               []sidebar.ChannelItem
 	// FinderItems is the merged list shown in the Ctrl+T finder. Initially
 	// contains only joined channels; the BrowseableChannelsLoadedMsg pipeline
 	// extends it with non-joined public channels in the background.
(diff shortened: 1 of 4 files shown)


gammons (commented):
Reviewed in a local worktree — build, `go test ./...`, and `go vet` all pass. The diagnosis (missing peer subscription + dropped batch `presence_change`) checks out, and the parser/test changes look good. A few findings:

1. **Dead call with an incorrect comment** (`cmd/slk/main.go:1864-1871`): the new `subscribeWorkspacePresence(wctx)` in `connectWorkspace` can't work — `Client.Connect` is REST-only (`auth.test`), and the WebSocket isn't dialed until `ConnMgr.Run` starts after `connectWorkspace` returns. The call always hits the `wsConn == nil` guard and silently fails into debuglog. The feature works via the `OnConnect` → `bootstrapPresenceAndDND` path; suggest removing this call (or correcting the comment).

2. **Unlocked read of `wctx.Channels`**: `workspacePresenceIDs` ranges over `wctx.Channels` from the `bootstrapPresenceAndDND` goroutine while the WS read goroutine can mutate it (`OnConversationOpened` appends, section/mute handlers write in place). The codebase already guards against exactly this — section rebootstrap runs synchronously on the read goroutine "to avoid racing wsCtx.Channels mutations" (main.go ~3189). Cheap fix: snapshot the IDs synchronously in `OnConnect` before spawning the goroutine.

3. **New DMs opened mid-session aren't subscribed until the next reconnect** — `OnConversationOpened` adds the item to `wctx.Channels` but never re-runs the subscription. May be acceptable scope for #69, but worth a one-liner there (which also runs on the read goroutine, sidestepping point 2 for that path) or a noted limitation.

Nit: ~60 lines of gofmt alignment churn in untouched declarations inflates the diff a bit.

PR #72, opened 2026-06-10
Restore the last-visited channel on startup

Frodotus:
## What

On startup, the active workspace now opens the **most-recently-visited channel** instead of always snapping to the sidebar's first entry — so a restart lands you back where you left off.

## How

The `channel_visits` table already persists a per-channel `last_visited` timestamp (and is loaded into `WorkspaceContext.LastVisitedByChannel` at connect). This wires that existing data into the startup path:

- `mostRecentlyVisitedChannel(visits)` picks the channel with the latest timestamp; it's passed on `WorkspaceReadyMsg.LastChannelID`.
- `reduceWorkspaceReady` (the `InitialActive` branch) opens `LastChannelID` when it's set and still present in the sidebar, otherwise falls back to `Channels[0]`.

Only the startup path changed; the existing mid-session workspace-switch restore (`lastChannelByTeam`) is untouched.

## Edge cases

- **First run / no recorded visit** → `LastChannelID` is empty → opens `Channels[0]` (unchanged behavior).
- **Last channel archived/left** → lookup finds no match → falls back to `Channels[0]`.

## Testing

`go test ./...` passes; added `TestWorkspaceReadyRestoresLastChannel` covering restore, the empty-visit fallback, and the stale-channel fallback.

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -2851,6 +2851,7 @@ func mostRecentlyVisitedChannel(visits map[string]int64) string {
 	}
 	return bestID
 }
+
 // rtmEventHandler bridges WebSocket events into bubbletea messages via p.Send()
 // and caches all incoming messages to the SQLite database.
 type rtmEventHandler struct {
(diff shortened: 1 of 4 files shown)


gammons:
One trivial fix: `cmd/slk/main.go` is missing a blank line between the new `mostRecentlyVisitedChannel` function's closing brace and the `// rtmEventHandler` doc comment, which is a `gofmt` violation that CI/`go fmt` will flag — otherwise the change looks correct and well-tested.

Frodotus:
Fixed in 03980a4 — added the blank line between `mostRecentlyVisitedChannel` and the `rtmEventHandler` doc comment. `gofmt -l` is clean on that change now.

gammons:
:hamburger:

PR #74, opened 2026-06-10
Open threads scrolled to the latest reply

### LINKED ISSUE #73: Threads open scrolled to the top instead of the latest reply
## Summary

Opening a thread scrolls to the top (parent / first reply) instead of the newest reply, unlike the Slack app where a thread opens at the latest message.

## Root cause

`thread.Model.SetThread` already selects the last reply (`m.selected = len(replies)-1`) intending to open at the bottom, but the viewport's snap-to-selection in `View()` only re-runs when the selection *index changes* (`!m.hasSnapped || m.snappedSelection != m.selected`). `SetThread` never resets the snap state, so opening a thread whose newest-reply index matches the previously-viewed thread's snapped selection — very common — skips the re-snap and leaves the viewport at the top.

`messages.Model.SetMessages` handles the analogous case by resetting `hasSnapped = false`; `SetThread` is missing that.

## Repro

1. Open a long thread → it opens at the bottom (snap runs the first time).
2. Scroll up, close it.
3. Open another long thread with a similar reply count → it opens at the top instead of the latest reply.

Frodotus:
Closes #73.

## Problem

Opening a thread scrolls to the top (parent / first reply) instead of the newest reply, unlike the Slack app. `SetThread` already selects the last reply (`m.selected = len(replies)-1`) intending to open at the bottom, but the viewport's snap-to-selection only re-runs when the selection *index changes*. Because `SetThread` never resets the snap state, opening a thread whose newest-reply index matches the previously-viewed thread's snapped selection (very common) skips the re-snap and leaves the viewport at the top.

## Fix

Reset `hasSnapped` in `SetThread` so the next `View()` re-snaps the viewport to the newest-reply selection — mirroring `messages.Model.SetMessages`, which already does exactly this with the same rationale.

## Testing

`go test ./...` passes; added `TestSetThreadOpensAtBottom`, which reproduces the matching-index scenario and asserts the viewport scrolls to the bottom on open (it fails without the one-line reset).

### DIFF
--- a/internal/ui/thread/model.go
+++ b/internal/ui/thread/model.go
@@ -309,6 +309,11 @@ func (m *Model) SetThread(parent messages.MessageItem, replies []messages.Messag
 	} else {
 		m.selected = 0
 	}
+	// Force the next View() to re-snap the viewport to the new selection.
+	// Without this, opening a thread whose newest-reply index matches the
+	// previously-viewed thread's snapped selection skips the snap and leaves
+	// the viewport scrolled to the top. Mirrors messages.Model.SetMessages.
+	m.hasSnapped = false
 	m.InvalidateCache()
 }
 
--- a/internal/ui/thread/open_bottom_test.go
+++ b/internal/ui/thread/open_bottom_test.go
@@ -1,0 +1,60 @@
+package thread
+
+import (
+	"fmt"
+	"testing"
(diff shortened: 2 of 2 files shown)


gammons:
LGTM, Thanks!

PR #77, opened 2026-06-10
Show name and avatar for bot messages

### LINKED ISSUE #76: Bot messages show no name or avatar
## Summary

Messages posted by bots/apps render with **no author name and no avatar** in the channel message pane. (The name is also missing wherever bot messages appear, e.g. thread replies.)

## Root cause

A `bot_message` carries no `user` field — only `bot_id`, `username`, and (sometimes) `icons`. slk's message conversion read only `m.User`, so bot messages ended up with `UserID == ""`:

- The avatar function looks up `AvatarURLs[""]` → miss → no avatar.
- The author name (`username`) was dropped entirely.

This affects *all* bots — even apps that have a bot user (`U…`) in `users.list`, because they still post as `bot_message` with an empty `user`. The bot's avatar may exist in the cache under its `U…` id, but the message only references `bot_id`.

Compounding it: messages already cached (with `UserID == ""`) stayed broken even after a fix, because the cached read path only read the stored `UserID` — though the cached `raw_json` does contain `bot_id`/`username`.

## Repro

1. Open a channel with a bot that posts via `bot_message` (e.g. a deploy/release bot).
2. Its messages show only a timestamp + text — no name, no avatar.

Frodotus:
Closes #76.

## Problem

Bot/app messages (`bot_message`) carry no `user` — only `bot_id`, `username`, and sometimes `icons`. The conversion read only `m.User`, so bot messages got `UserID == ""`: the avatar lookup missed and the author name was dropped. This hit every bot, since even apps with a bot user post as `bot_message` with an empty `user`.

## Fix

- **Identity:** key bot messages on `bot_id` and use the message's `username`, in all three fetch paths (new `messageAuthor` helper) and the live WebSocket path (`OnMessage` now receives `bot_id`/`username`, threaded through the event struct + dispatch).
- **Avatar:** a new `userResolver.RequestBot` resolves the icon via **`bots.info`** once per bot (deduped, persisted) and preloads it via `avatars.Preload(bot_id, …)`; `AvatarURLs[bot_id]` is then populated on the resolve/connect path — mirroring the `users.info` path for humans. Added `GetBotInfo` to the client.
- **Cached rows:** messages cached with `UserID == ""` before this change are re-keyed by parsing `bot_id`/`username` out of `raw_json` in `enrichCachedRow`, so existing cached bot messages render without a cache wipe.

## Testing

`go test ./...` passes. Added `TestMessageAuthorBotIdentity` (fetch-path identity) and extended `TestDispatchWebSocketBotMessageEvent` to assert `bot_id`/`username` flow through the WS dispatch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 bin/
 dist/
+/slk
 *.db
 *.db-journal
 .superpowers/
--- a/cmd/slk/bot_identity_test.go
+++ b/cmd/slk/bot_identity_test.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/slack-go/slack"
+)
+
+// TestMessageAuthorBotIdentity is the headline of the bot-avatar fix: a
+// bot_message (no user, only bot_id + username) is keyed on the bot_id and
+// uses the message's username, so a name and avatar can attach — while
+// human messages keep resolving by user ID.
+func TestMessageAuthorBotIdentity(t *testing.T) {
+	userNames := map[string]string{}
+
+	// Bot message: empty user, bot_id + username present.
+	bot := slack.Message{Msg: slack.Msg{User: "", BotID: "B123", Username: "Deploybot"}}
+	uid, name := messageAuthor(bot, userNames, nil, nil)
+	if uid != "B123" {
+		t.Errorf("bot author ID: want B123, got %q", uid)
+	}
+	if name != "Deploybot" {
+		t.Errorf("bot author name: want Deploybot, got %q", name)
+	}
+
+	// Human message: user present, name pre-cached (so no db/resolver needed).
+	userNames["U1"] = "Alice"
+	human := slack.Message{Msg: slack.Msg{User: "U1"}}
+	uid2, name2 := messageAuthor(human, userNames, nil, nil)
+	if uid2 != "U1" {
+		t.Errorf("human author ID: want U1, got %q", uid2)
+	}
+	if name2 != "Alice" {
+		t.Errorf("human author name: want Alice, got %q", name2)
+	}
+}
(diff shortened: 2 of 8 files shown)


gammons:
Reviewed in a worktree — build passes, `go vet` clean, all tests green (including the two new bot-identity tests). **Approve with minor nits.**

Well-targeted fix: correct root cause (bot_message has empty `user`, only `bot_id`/`username`), consistently applied across all three fetch paths, the live WS path, and the cached-row backfill. `RequestBot` faithfully mirrors `Request` and the avatar pipeline wires through unchanged since cache + MessageItem are keyed on the same `bot_id`. The `enrichCachedRow` re-keying for pre-fix rows (avoiding a cache wipe) is a nice touch. Tests are meaningful, not signature-padding.

Nits (non-blocking):
- `bestBotIcon` comment says "largest reasonable icon," but the fallback order (72, 48, 132, 36, 230) deliberately prefers 72 over the larger sizes. The ordering is fine; the comment should reflect the actual intent.
- `ShouldNotify` uses the original `userID` (empty for bots) rather than `authorID` — almost certainly intentional (a bot cant be you), but worth a one-line comment so it is not "fixed" later.
- PR description says the avatar is stored into `AvatarURLs[bot_id]`; precisely, `RequestBot` calls `avatars.Preload(botID, ...)` and `AvatarURLs` is populated on the `UserResolvedMsg`/connect path. Functionally correct, just imprecise wording.

Frodotus:
Thanks for the review! Addressed the nits in f387eb9 (non-functional, comments + wording only):

- **`bestBotIcon` comment** — rewritten to match the actual intent: prefer the 72px icon (sharp enough, almost always present) and fall back through the other sizes *by availability*, not strictly by pixel size.
- **`ShouldNotify` userID** — added a one-line comment at the call site noting the empty `userID` for bots is intentional: the self-message suppression keys on the human sender, and a bot (`userID == ""`, `authorID == botID`) can never be "you".
- **PR description** — corrected the avatar wording: `RequestBot` calls `avatars.Preload(bot_id, …)`, and `AvatarURLs[bot_id]` is populated on the resolve/connect path (not stored directly by `RequestBot`).

PR #82, opened 2026-06-11
fix(app): switch to native clipboard library

### LINKED ISSUE #80: Permalinks are not working on ghostty + macosx
When I press Y it shows `Copied permalink` at the application footer, but when I try to paste it anywhere else it does not work.

gcgbarbosa:
Replaces `tea.SetClipboard` with `golang.design/x/clipboard` . Fixes copying failures on macOS (especially inside tmux environments). Works across macOS, Linux, and Windows.

Fixes #80

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -909,6 +909,9 @@ func (a *App) copyPermalinkOfSelected() tea.Cmd {
 			return nil
 		}
 
+		if !a.clipboardAvailable {
+			return statusbar.PermalinkCopyFailedMsg{}
+		}
 		_ = clipboard.Write(clipboard.FmtText, []byte(url))
 
 		return statusbar.PermalinkCopiedMsg{}
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -151,6 +151,11 @@ type App struct {
 	// clipboard contents. Tests inject fakes via SetClipboardReader.
 	clipboardRead clipboardReader
 
+	// clipboardWrite is the function used by copyPermalink and drag-copy
+	// to write OS clipboard contents. Tests inject fakes via
+	// SetClipboardWriter.
+	clipboardWrite clipboardWriter
+
 	// threads is the App's ThreadService collaborator (fetch / mark /
 	// reply / list-fetch + parent-channel last-read lookup for the
 	// unread boundary). See internal/ui/services.go. Defaulted to a
@@ -422,6 +427,7 @@ func NewApp() *App {
 		browserOpener:        openURLCmd,
 		navHistory:           newNavHistoryStore(),
 		clipboardRead:        defaultClipboardReader,
+		clipboardWrite:       defaultClipboardWriter,
 	}
 	app.editing = newEditController()
 	// typing tracker is referenced by typingOut so it must exist first;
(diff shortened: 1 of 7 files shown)


gammons:
Thanks for this — switching to `golang.design/x/clipboard` is the right call and matches how the rest of the app already handles the clipboard (smart-paste, image attach). One blocking issue plus minor cleanup:

**Blocking: unguarded `clipboard.Write` can panic the app.** The library requires a successful `clipboard.Init()` before any `Write`; its docs note that otherwise Read/Write "may result in an unrecoverable panic." Everywhere else in the codebase the clipboard is gated behind `a.clipboardAvailable` (set from the `clipboard.Init()` result in `cmd/slk/main.go`), e.g. `smartPaste`. This new path is the only unguarded call, so on environments where init failed (Wayland without wl-clipboard, headless/SSH, missing libx11) pressing `Y` would crash instead of no-op.

**Minor: false success toast.** `_ = clipboard.Write(...)` discards the result and then always returns `PermalinkCopiedMsg{}`, so a failed write still shows "Copied permalink" — basically the original #80 symptom relocated. Guarding on `clipboardAvailable` fixes both:

```go
if !a.clipboardAvailable {
    return nil
}
_ = clipboard.Write(clipboard.FmtText, []byte(url))
return statusbar.PermalinkCopiedMsg{}
```

**Cosmetic:** `gofmt -l` flags the file — a double blank line in the imports, trailing whitespace on two blank lines, and a stray blank line before `openLinksOfSelected`. A quick `gofmt -w internal/ui/app.go` cleans it up.

gcgbarbosa:
should be good now @gammons

gammons:
Thanks for the updates — the `clipboardAvailable` guard is the right fix, and extending it to the mouse-selection copy in `drag.go` is a nice catch.

One blocker though: CI will be red. `go test ./...` fails 5 tests that pass on `main`:

```
TestCopyPermalink_FromMessagesPane    (got PermalinkCopyFailedMsg, want PermalinkCopiedMsg)
TestCopyPermalink_FromThreadPane
TestCopyPermalink_ShiftYTriggersCopy
TestApp_DragInMessagesEmitsClipboardAndToast
TestDrag_MotionCoalescing_ReleaseFlushesPending
```

Root cause: tests build the App without calling `clipboard.Init()`, so `clipboardAvailable` is `false` and the new guard short-circuits the success path to the failure message. Clipboard *reads* are injectable (`clipboardRead` / `SetClipboardReader`) but the new *writes* call package-level `clipboard.Write` directly, so tests can't simulate a successful copy — and flipping `clipboardAvailable = true` would then hit the real `clipboard.Write`, which can panic in headless CI.

Suggest mirroring the read seam: add a `clipboardWrite` field (default `clipboard.Write`) + `SetClipboardWriter` hook, use it in `app.go` and `drag.go`, then update the 5 tests to inject a fake writer (and the two drag tests to stop expecting `tea.SetClipboard`).

Minor: `gofmt -l` still flags `internal/ui/app.go` (trailing whitespace on the two tab-only blank lines + a double blank line before `openLinksOfSelected`).

gammons:
thanks @gcgbarbosa !

PR #84, opened 2026-06-11
Add keep_focus_on_list: stay on the list when selecting a channel or opening a thread

### LINKED ISSUE #83: Feature proposal: keep_focus_on_list — stay on the list when selecting a channel or opening a thread
## Motivation

Picking a channel from the sidebar — or opening a thread — currently moves keyboard focus into the content pane. That's ideal when you want to read/react/reply immediately, but it makes *browsing* awkward: to skim several channels (or several threads) you select one, then have to `Tab`/`h` back to the list before selecting the next. A "stay on the list" mode lets you walk the list with `j/k` + `Enter`, loading each selection's content, without leaving the list each time.

## Proposed design

**Config** — a new `[general]` bool, default `false` (today's behavior, opt-in only):
```toml
[general]
keep_focus_on_list = true
```

**Behavior when `true`** — selecting still loads the content, but keyboard focus stays on the list:
- sidebar → channel: focus stays on the **sidebar**
- channel message → its thread: thread panel opens, focus stays on the **message list**
- threads view → thread: thread opens, focus stays on the **threads list**

Step into the content when you actually want to interact, via the existing focus keys (`Tab` / `l` / `→`). `Enter` only (re)loads in this mode.

**Scope notes** — the flag governs the three *list→content* focus jumps only. Switching *which* list you browse is unchanged: activating the Threads view still lands you on the threads list (that list is what you browse there), and mouse click-to-focus-a-pane is untouched.

**Implementation** — gate the three `focusedPanel` assignments (the channel-select reducer + the two thread-open paths) behind the flag, and wire `config.General.KeepFocusOnList` through to the App (mirroring the other appearance/general bools). Small and additive; the default preserves current behavior for anyone who doesn't opt in.

Frodotus:
Closes #83.

## What

A new `[general]` bool, `keep_focus_on_list` (default `false`). When enabled, selecting a channel or opening a thread loads the content but leaves keyboard focus on the list, so you can browse with `j/k` + `Enter` without tabbing back each time.

```toml
[general]
keep_focus_on_list = true
```

## Behavior (when `true`)

Selecting still loads the content; only the focus jump is suppressed:

- sidebar → channel: focus stays on the **sidebar**
- channel message → its thread: thread opens, focus stays on the **message list**
- threads view → thread: thread opens, focus stays on the **threads list**

Step into the content via the existing focus keys (`Tab` / `l` / `→`). Default (`false`) is the long-standing focus-follows-selection behavior, so nobody is affected unless they opt in.

## Implementation

- Gates the three selection-driven `focusedPanel` assignments — `reduceChannelSelected` (channel pick), `handleEnter` (threads-view Enter), and `openThreadPanel` (message → thread) — behind `!a.keepFocusOnList`.
- Wires `config.General.KeepFocusOnList` → `app.SetKeepFocusOnList(...)`, mirroring the other `[general]`/`[appearance]` bools.
- **Intentionally not gated:** `ThreadsViewActivatedMsg` still focuses the threads list — that list *is* what you browse in the threads view, so landing on it is the on-list behavior (commented inline). Mouse click-to-focus-a-pane is also untouched.

## Tests

Three tests, each driving a real production path and asserting default-vs-flag:
- channel select via the `ChannelSelectedMsg` reducer (`Update`)
- message → thread via the real `handleEnter` keystroke
- threads view → thread via the real `handleEnter` keystroke

Full suite + `go vet` green. Wiki `Configuration.md` updated ("Browsing without losing focus").

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -542,6 +542,7 @@ func run() error {
 	app := ui.NewApp()
 	app.SetHelpFooter(versionpkg.ModalFooter(version))
 	app.SetClipboardAvailable(clipboardOK)
+	app.SetKeepFocusOnList(cfg.General.KeepFocusOnList)
 	if useWaylandClipboard {
 		app.SetClipboardReader(ui.WaylandClipboardReader())
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -67,6 +67,12 @@ type General struct {
 	// instead of the config-glob [sections.*] system. Pointer so we
 	// can distinguish "unset" (default true) from explicit false.
 	UseSlackSections *bool `toml:"use_slack_sections"`
+	// KeepFocusOnList keeps keyboard focus on the list when you select a
+	// channel or open a thread, instead of moving focus into the content
+	// pane. Lets you browse with j/k + Enter without tabbing back to the
+	// list each time; step into the content with Tab / l / →. Default
+	// false (focus follows selection — the long-standing behavior).
+	KeepFocusOnList bool `toml:"keep_focus_on_list"`
 }
 
 type Appearance struct {
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -147,6 +147,12 @@ type App struct {
 	// clipboard.Init(). When false, Ctrl+V smart-paste is a no-op.
 	clipboardAvailable bool
 
+	// keepFocusOnList, when true, keeps keyboard focus on the list
+	// (sidebar / threads list / message list) after selecting a channel
+	// or opening a thread, rather than moving focus into the content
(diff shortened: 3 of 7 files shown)


gammons:
hi @Frodotus curious why you're closing your PRs!

Frodotus:
> hi @Frodotus curious why you're closing your PRs!

Hi! Sorry for the confusion — no slight intended! I opened those while trying to get slk working as my daily driver, but I'm not really a Go person. In the end I built my own client in Python since it's what I know — that way I can quickly tweak things to fit my daily-driver needs instead of it turning into a Go learning project. Really appreciate slk though, and the effort you've put into it!

PR #90, opened 2026-06-30
Add flake.nix

jennydaman:
A `flake.nix` makes it easy for Nix and [NixOS](https://nixos.org/) users to install `slk`.

However, there are some considerations for accepting this PR:

- version string must be hard-coded into `flake.nix`
- `vendorHash` must be updated manually when `go.sum` is modified


gammons:
thanks!

Issue #89, opened 2026-06-24
Feature Request: Option to hide workspace minibar


mblarsen:
The workspace minibar takes up 4 chars width which is a lot on a small screen.
Many users do not have more than one workspace or don't twitch that frequently.

Suggestions:

1. Option to hide/toggle minibar
2. Show current the two letter workspace ID in status bar as well

<img width="92" height="74" alt="Image" src="https://github.com/user-attachments/assets/27ca31ea-8ef7-4dd8-a17b-08f983cdc422" />

hkdd:
I added a pull request for this to the author (gammons).  If you want to test it now, you can pull from my branch.  adds the ctrl-l (and ctrl-\ fallback) to toggle the workspaces bar.  

https://github.com/hkdd/slk/tree/fix/89-hide-workspaces

gammons:
likely going into next release

PR #95, opened 2026-07-07
From issue #89, add a keybind to toggle the workspaces sidebar.

hkdd:
### Summary of Changes:

  1. Defined Keybindings: Added  ToggleWorkspaces  to  KeyMap  inside keys.go, binding it to  ctrl+l  (and  ctrl+\  as an alias).
  2. Added State Field: Extended  App  inside app.go to track  workspacesVisible  state and default it to  true  on initialization.
  3. Implemented Toggle Action: Created  ToggleWorkspaces()  method in app.go.
  4. Wired Normal Mode Handler: Added a check for key match against  a.keys.ToggleWorkspaces  in mode_normal.go.
  5. Adjusted Layout & Assembly:
      • In  App.View() , the layout width for the rail defaults to  0  when  workspacesVisible  is  false , and rendering of the rail is skipped.
      • Updated the  buildViewPanels()  test helper in view_composite_test.go to match.
  6. Added Unit Tests: Added  TestAppToggleWorkspaces  to app_test.go verifying toggling and visibility state changes.

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -106,8 +106,9 @@ type App struct {
 	// State
 	mode           Mode
 	focusedPanel   Panel
-	sidebarVisible bool
-	threadVisible  bool
+	sidebarVisible    bool
+	workspacesVisible bool
+	threadVisible     bool
 	view           View
 	width          int
 	height         int
@@ -459,6 +460,7 @@ func NewApp() *App {
 		wins:                 wins,
 		focusedWin:           rootWin,
 		sidebarVisible:       true,
+		workspacesVisible:    true,
 		view:                 ViewChannels,
 		keys:                 DefaultKeyMap(),
 		selfSend:             newSelfSendDedup(),
@@ -1610,6 +1612,11 @@ func (a *App) ToggleSidebar() {
 	}
 }
 
+func (a *App) ToggleWorkspaces() {
+	a.clearSelections()
+	a.workspacesVisible = !a.workspacesVisible
+}
+
 func (a *App) ToggleThread() {
 	a.clearSelections()
 	if a.threadVisible {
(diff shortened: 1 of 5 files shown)


hkdd:
For the sake of full disclosure:  this was 100% created with Google Antigravity using Gemini 3.5 Flash.

gammons:
Thanks @hkdd — this is close. Two things worth addressing before merge (a real edge-case bug + a test gap):

**1. `windowBounds()` still reserves the rail when workspaces are hidden**

`View()` now zeroes the rail width when `!workspacesVisible`, but `windowBounds()` (`internal/ui/windows.go:51`) still passes the full width unconditionally:

```go
frame := a.layout.Compute(a.width, a.height, a.workspaceRail.Width(), ...)
```

So the rail width now has two sources of truth. When workspaces are hidden **and** split windows are in use:
- the window tree subdivides a message region computed as if the rail were still present (narrower than what `View()` actually renders), and
- `panelLayout.Compute` has a side effect (`l.railWidth = railWidth`, `panellayout.go:128`), so calling `windowBounds()` while hidden transiently clobbers the mouse hit-test bands until the next `View()`. The comment at `windows.go:48` ("Compute is deterministic for unchanged inputs") no longer holds, because the same state now feeds two different `railWidth` inputs.

Suggest extracting a single helper and using it in `View()`, `windowBounds()`, and the `buildViewPanels` test helper (which currently hand-copies the same conditional):

```go
func (a *App) effectiveRailWidth() int {
    if !a.workspacesVisible {
        return 0
    }
    return a.workspaceRail.Width()
}
```

**2. Test only covers the boolean flip**

`TestAppToggleWorkspaces` asserts `workspacesVisible` toggles, but not the actual behavior we care about — that hiding the rail removes it from the rendered layout. A test that renders with `workspacesVisible=false` and asserts the rail panel is absent / `frame.RailWidth == 0` would guard the View/layout wiring, which is where the real risk is. The `buildViewPanels` helper already supports this.

Minor: the `ctrl+l` binding collides with the terminal redraw convention, and the `ctrl+\` alias is SIGQUIT + not shown in help — recommend a single, out-of-the-way binding like `ctrl+g` instead. Also run `gofmt` on `app.go`/`keys.go` (the new longer identifiers left the struct + KeyMap literal misaligned).

gammons:
closing for now.  please re-open if you'd like to resubmit!

PR #96, opened 2026-07-08
notify: add notify_command + status_command hooks, and respect muted conversations

abv:
## What

Three related additions to slk's notification path (opt-in; default behavior unchanged):

1. **`notify_command`** (`[notifications]`) — runs *instead of* the built-in OS notification for any message that would notify (DM / mention / keyword), via `sh -c` with `$SLK_TITLE` / `$SLK_BODY`.
2. **`status_command`** (`[notifications]`) — runs on every unread-state change (a message arrives *and* a channel is read) so an external surface can mirror unread state. Gets `$SLK_UNREAD`, `$SLK_OTHER_UNREAD`, `$SLK_WORKSPACE`, `$SLK_TITLE`. Because it fires on reads too, it can clear a status as well as set one.
3. **Respect muted conversations** — `ShouldNotify` now suppresses a message from a muted channel/DM, matching Slack. Previously `MuteStore` only affected unread counts and the sidebar, not notifications.

(1) and (2) are the first commit; (3) is a separate commit and can be split into its own PR if you'd prefer.

## Why

slk already computes exactly the state these need — the notification triggers, the unread counts behind the `slk SW (3) +1` window title, and the mute set — but there's no way to route any of it anywhere but the built-in OS notification / terminal title, and muted channels still notified. These hooks let people wire slk into `terminal-notifier`, `mako`, a status bar, or a terminal multiplexer's sidebar, with no slk-side knowledge of any of them. (I route both into [herdr](https://github.com/ogulcancelik/herdr): an in-multiplexer toast + a self-clearing sidebar status.)

## How

- **`internal/notify`**: `Notifier` gains an optional command run in place of `beeep`; new `StatusReporter` runs the status command. Both pass values through the **environment** (never string-interpolated), so message text / workspace names can't inject shell syntax. `ShouldNotify` gains an `IsMuted` short-circuit.
- **`internal/config`**: two opt-in `Notifications` fields; empty = current behavior.
- **`internal/ui`**: `App` gains a `StatusReportFunc` callback (`SetStatusReporter`), invoked from the existing `notifyReadStateChanged` hook (the same spot that recomputes the window title).
- **`cmd/slk`**: wires both from config (the status command runs off the UI goroutine) and passes mute state into `ShouldNotify`.

## Tests

`go test -race ./...` and `go vet ./...` pass. New tests cover: the command runs with the expected env; disabled / empty is a no-op; nil-safe; an **injection** test (env-passed values can't spawn a second command); and mute suppression.

## Config

Both hooks are opt-in and documented in `wiki/Configuration.md`:

```toml
[notifications]
notify_command = 'terminal-notifier -title "$SLK_TITLE" -message "$SLK_BODY"'
status_command = 'my-statusbar --slack-unread "$SLK_UNREAD"'
```

With neither set, behavior is unchanged.

## Example: wiring into herdr

For reference, how these are used in practice — routed into [herdr](https://github.com/ogulcancelik/herdr), which exports `$HERDR_PANE_ID` into each pane — in `~/.config/slk/config.toml`:

```toml
[notifications]
on_keyword = ["deploy"]
# DM / mention / keyword -> an in-herdr toast
notify_command = 'herdr notification show "$SLK_TITLE" --body "$SLK_BODY" --sound request'
# unread state -> a sidebar status circle that self-clears on read
status_command = 'if [ "$SLK_UNREAD" -gt 0 ] || [ "$SLK_OTHER_UNREAD" -gt 0 ]; then herdr pane report-agent "$HERDR_PANE_ID" --source slk --agent slk --state blocked; else herdr pane release-agent "$HERDR_PANE_ID" --source slk --agent slk; fi'
```

---

*AI-assisted, per the contributing guidelines: authored with **Claude Opus 4.8** at high thinking effort, tests-first and human-reviewed. I understand the diff; it builds and passes `go vet ./...` and `go test -race ./...`. Commits carry `Co-Authored-By` accordingly.*


gammons:
@abv Thanks for this — reviewed the full diff, and it's a well-crafted, high-quality PR. Opt-in, default behavior unchanged, the env-passing security model is correct and explicitly tested, docs updated, and the commit split is clean. `go build`, `go vet`, and `go test ./internal/notify/...` all pass locally. Test coverage (injection, nil-safety, disabled-is-noop, env plumbing) is genuinely solid.

One thing I'd like addressed before merge, plus some minor notes:

**1. `status_command` out-of-order-completion hazard (substantive)**

In `cmd/slk/main.go` the reporter is dispatched per-call as `go sr.Report(...)`. `notifyReadStateChanged` runs on the single-threaded bubbletea Update loop, so the *reads* are consistent — but the goroutine dispatches are not serialized with each other. Under a burst (bulk catch-up marking several channels read, or a flurry of incoming messages), multiple overlapping `sh -c` subprocesses race, and if state goes A→B the subprocess for A can finish after B, leaving the external surface pinned to a stale count. For a feature whose whole point is to *mirror* current unread state, a stale final write is a correctness bug.

Suggested fix: serialize through a single worker that coalesces — one goroutine, a latest-only channel, always run the newest and drop intermediate states. That also bounds subprocess fan-out under bursts.

**Minor / optional:**

- **`sh -c` is POSIX-only** — both hook paths hardcode `sh`, so they silently no-op on Windows (where beeep otherwise works). Fine given slk is a terminal client, just worth documenting as a limitation.
- **Silent hook failures** — `go ...Notify(...)` and `go sr.Report(...)` discard the returned error, so a misconfigured command fails invisibly. Reasonable for a TUI, but consider routing to `debuglog` for diagnosability.
- **Mute semantics change** — placing the `IsMuted` short-circuit before the mention/keyword checks means an @mention/keyword in a muted channel that previously notified now stays silent. Documented in the commit as intentional ("matching Slack") and I think it's the right call; just flag it as a behavior change for existing users since it's a separable commit.
- Nit: `NewStatusReporter` returning nil + `Report` being nil-safe + the `if sr != nil` guard in main.go is three layers of belt-and-suspenders. Harmless.

Verdict: approve with the one requested change (#1). Everything else is optional polish. The security discipline and test coverage stand out — nice work. Happy to sketch the coalescing worker if useful.

abv:
Thanks for the thorough review! Pushed 1416767 addressing all of it.

**1. Out-of-order-completion hazard — fixed as suggested.** `StatusReporter` now owns a single worker goroutine fed by a capacity-1 newest-wins mailbox. main.go wires `sr.Enqueue` straight into `SetStatusReporter`: it never blocks the UI goroutine, runs are serialized so a slow subprocess can never overlap — and finish after — a newer state's run, and intermediate states coalesce away under a burst (which also bounds subprocess fan-out to one at a time). Two new deterministic tests pin both properties:

- a gate-file test: the first run blocks on a file, a burst of 8 states is enqueued while it's provably in-flight, and the executed states must be exactly `[first, newest]` — no overlap, no intermediate runs;
- a 50-state burst test asserting executed states form a strictly in-order subsequence ending in the final state (fails under the old `go sr.Report(...)` dispatch).

**Minor notes, also taken:**

- **Silent hook failures** → both hook paths now log through a new `[notify]` debuglog category; Configuration.md points at `SLK_DEBUG=1` + grepping `[notify]` in slk-debug.log as the way to diagnose a misbehaving command.
- **`sh -c` POSIX-only** → documented as a limitation in Configuration.md (hooks unavailable on Windows; the built-in OS notification still works there).
- **Mute semantics** → flagged explicitly in Configuration.md as a behavior change (a mention/keyword in a muted channel no longer notifies). Still happy to split that commit into its own PR if you'd rather land it separately — just say the word.
- **Nil-safety nit** → acknowledged; left as is since `Enqueue` inherits the same guard and each layer stays independently safe.

`go build`, `go vet ./...`, and `go test -race ./...` all pass, and the notify package passes repeated `-count` runs.

*Same workflow as the rest of the PR: AI-assisted, human-reviewed; the new commit carries Co-Authored-By.*

PR #99, opened 2026-07-16
Resolve usergroup mentions to @handles and support tagging groups

rfist:
## What this fixes

When someone tags a team in Slack (like `@platform-team`), slk shows the raw wire token instead of the team name:

> dear `<!subteam^S0TESTGRP01>` please delete the recurring event

You can't tell who's being addressed without opening the official client. The same tokens leak into search results, thread exports, and desktop notifications. And there's no way to tag a team from slk at all — typing `@backend-teams` sends plain text that pings nobody.

## What changes for users

- **Team mentions render as names**: `<!subteam^S0TESTGRP01>` shows as **@platform-team** everywhere — message pane, thread panel, search snippets, markdown export, and desktop notifications.
- **You can tag teams**: the `@` autocomplete in the composer now lists your workspace's user groups alongside people. Selecting one sends a real mention — members get notified, same as tagging from the official client.
- **Zero configuration**: group names load automatically per workspace on connect. If the fetch fails, mentions degrade to a readable `@group` placeholder instead of the raw token.
- **Broadcast mentions render too**: `<!channel>`, `<!here>`, and `<!everyone>` were also leaking as raw tokens in the message pane and markdown export (the snippet flattener and notifications already handled them) — they now render as styled **@channel** / **@here** / **@everyone** everywhere.

## How it works

- Each workspace fetches its user groups via `usergroups.list` in the background on connect (works with the xoxc browser token slk already uses).
- The ID → handle map is workspace-scoped, carried in `WorkspaceContext` and propagated to the UI via a `UserGroupsLoadedMsg` (guarded by team ID, mirroring the existing `CustomEmoji`/`UserNames` flow). Groups from one workspace never leak into another's picker or send-time translation.
- `internal/usergroups` holds pure helpers (`Copy`, `Display`); rendering falls back embedded label → workspace map → `@group`.
- Outgoing mentions are translated `@handle` → `<!subteam^SID|@handle>` (labeled form, so the optimistic echo renders without waiting for a lookup).

Test coverage: all render paths, picker entries, membership filtering, send translation, and workspace collision/leak regressions (inactive workspace's groups must not appear in the active picker).

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -165,7 +165,20 @@ type WorkspaceContext struct {
 	UserID        string
 	UnresolvedDMs []UnresolvedDM
 	CustomEmoji   map[string]string // emoji name -> URL or "alias:target"
-	UserGroups    map[string]string // usergroup ID -> handle
+	// userGroups holds this workspace's usergroup ID -> handle map.
+	// Access it via UserGroups/SetUserGroups, never directly.
+	//
+	// atomic.Pointer (not a plain map) because the write happens on the
+	// background usergroups.list fetch goroutine while reads happen on
+	// the bubbletea Update/cmd goroutines (workspace switch, search) and
+	// on the RTM event loop (notification body stripping). This is where
+	// UserGroups differs from the CustomEmoji field it otherwise mirrors:
+	// CustomEmoji is only ever consumed through the tea message loop, so
+	// its single background write is never read cross-goroutine.
+	//
+	// The stored map is published once and never mutated afterwards, so
+	// readers need no further synchronization.
+	userGroups atomic.Pointer[map[string]string]
 	// Self presence and DND state for this workspace. Populated on connect
 	// and updated by manual_presence_change / dnd_updated WS events plus
 	// optimistic writes from the presence menu.
@@ -192,6 +205,22 @@ type WorkspaceContext struct {
 	// in connectWorkspace alongside UserResolver (it depends on the
 	// resolver to trigger external-user lookups for newly-seen IDs).
 	Membership *membership.Manager
+}
+
+// UserGroups returns this workspace's usergroup ID -> handle map, or an
+// empty map before usergroups.list has returned. Safe to call from any
+// goroutine; the result must be treated as read-only.
+func (w *WorkspaceContext) UserGroups() map[string]string {
+	if m := w.userGroups.Load(); m != nil {
+		return *m
+	}
+	return map[string]string{}
+}
+
+// SetUserGroups publishes a usergroup ID -> handle map for this
+// workspace. The caller must not mutate the map afterwards.
+func (w *WorkspaceContext) SetUserGroups(groups map[string]string) {
+	w.userGroups.Store(&groups)
 }
 
 // workspaceRouter holds the program-wide "active workspace" pointer.
@@ -1529,7 +1558,7 @@ func run() error {
 			ExternalUsers:    external,
 			UserID:           wctx.UserID,
 			CustomEmoji:      wctx.CustomEmoji,
-			UserGroups:       wctx.UserGroups,
+			UserGroups:       wctx.UserGroups(),
 			SectionsProvider: sectionsProviderAdapter{store: wctx.SectionStore},
 		}
 	})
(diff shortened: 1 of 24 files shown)


gammons:
## Code Review

@rfist Solid, well-structured PR — clean separation via the pure `internal/usergroups` helpers, a workspace-scoped map threaded consistently through every render/compose surface, backward-compatible wrapper functions, and genuinely thorough tests (including the workspace-leak regression). Build passes and all affected tests pass under `-race`; gofmt/vet clean.

One real concurrency issue **worth fixing before merge**; the rest are minor.

### 1. [Concurrency — worth fixing] Data race on `wctx.UserGroups`
- `cmd/slk/main.go:1720` writes `wctx.UserGroups = byID` from the background fetch goroutine.
- `cmd/slk/main.go:3328` reads `h.wsCtx.UserGroups` from the RTM `OnMessage` handler goroutine.

These are different goroutines sharing the same `*WorkspaceContext` with no synchronization — a data race per Go's memory model. This is the one place the PR diverges from the `CustomEmoji` precedent it models itself on: `CustomEmoji` is written the same way (main.go:1702) but is *only* consumed through the tea message loop (`p.Send`), never read cross-goroutine. `UserGroups` is additionally read directly on the WS event loop for notifications. Other `wsCtx` fields read in `OnMessage` (e.g. `DNDEnabled`) are written on the same WS loop, so they're safe — `UserGroups` is uniquely written from a *different* goroutine.

Practical impact is low (a single one-time pointer-sized write, small window; worst realistic outcome is one notification using a stale/empty map), but `go test -race` on an integration path would flag it. Simplest fixes: have `OnMessage` source the map from a handler field updated on the WS loop via the existing `UserGroupsLoadedMsg` path, or use an `atomic.Pointer`/mutex snapshot.

### 2. [Minor] Nondeterministic tie-break in `TranslateMentionsForSend` (`internal/ui/compose/model.go:940`)
Entries are sorted by name length descending with `sort.Slice` (unstable) over map-iteration order. If a username and a usergroup handle are the same length (e.g. both `@design`), which wire form wins is nondeterministic. Consider a deterministic secondary tie-break or documenting precedence.

### 3. [Nit] `usergroupHandles` name fallback can produce handles with spaces (`cmd/slk/main.go:1779`)
When `g.Handle == ""` it falls back to `g.Name` ("Platform Team"). A space-containing handle won't round-trip cleanly through the `@`-token composer autocomplete or the word-boundary send translation. Slack handles are normally space-free so risk is low; consider slugifying or skipping when `Handle` is empty.

### 4. [Nit] Inconsistent `SetUserGroups` implementations
`threadsview.Model.SetUserGroups` normalizes nil→`{}` and guards with `stringMapsEqual` before marking dirty; `messages.Model` and `thread.Model` unconditionally invalidate the cache on every call. Since `App.SetUserGroups` fires on every workspace switch, those models bust their render caches even for identical sets. Minor perf only — match the guard or confirm it's intentional.

---

None of the nits are blocking. Recommend addressing **#1** before merge (or confirming the team accepts it as a known benign race consistent with existing patterns).

rfist:
Thanks — all four addressed in de69a90.

**1. Data race on `wctx.UserGroups`** — fixed. The field is now an unexported `atomic.Pointer[map[string]string]` behind `UserGroups()` / `SetUserGroups()` accessors.

One thing worth flagging: `OnMessage` isn't the only racy read. The workspace-switch cmd (`main.go:1561`) and the search cmd (`main.go:3045`) also read the map from bubbletea cmd goroutines, so they race with the same background write. That's why I went with the atomic pointer over your other suggestion of an `rtmEventHandler` field fed by the WS loop — that fixes the notification read but leaves those two. The map is published once and never mutated afterwards, so a pointer swap is enough; no mutex, no copy on read. `UserGroups()` returns an empty non-nil map before the fetch lands, preserving the `make(map[string]string)` zero value the constructor used to set.

Your read on the `CustomEmoji` precedent was right and I've written it into the field comment, so the next person doesn't "simplify" this back to a plain map for symmetry.

**2. Nondeterministic tie-break** — fixed. Added an explicit `rank` to the entry struct (special < user < usergroup < channel) and a full chain: length → rank → name → replacement. Rather than just documenting precedence I made it deliberate: for a name shared by a person and a group, the person wins, since the failure mode in that direction is one wrong DM-ish ping instead of notifying an entire group.

**3. Handle fallback with spaces** — fixed via `slugifyHandle`: lowercase, runs of non-handle characters collapse to `-`, `_` and `.` preserved. Groups that slugify to nothing are skipped rather than registered as unmentionable, per your "or skipping" suggestion.

**4. Inconsistent `SetUserGroups`** — not intentional, fixed. Added `usergroups.Equal` (nil and empty compare equal) and used it in `messages`, `thread`, and `threadsview`, replacing threadsview's local `stringMapsEqual` guard.

New tests: `cmd/slk/usergroups_test.go` (slugify table, skip rules, concurrent accessor access that fails under `-race` if this regresses to a plain map), compose collision determinism, a `messages` cache-guard test, and an `usergroups.Equal` table. `go build`, `go vet`, `gofmt` clean; `go test ./... -race` fully green.

Manual testing: group name rendering, `@` autocomplete, and the `@here`/`@channel`/`@everyone` aliases all verified. The notification body and the multi-workspace switch weren't manually exercised this round — single workspace on hand — so those rest on the unit tests plus the leak regression test.

PR #100, opened 2026-07-16
Auto-detect sixel for iTerm2 image rendering

rfist:
On iTerm2, `Detect`'s auto path falls through to halfblock, so inline images render as a blurry 2-pixels-per-cell mosaic. iTerm2 has no working kitty graphics implementation (its partial support fails the startup probe — see the note on `ProbeKittyGraphics`), but it does support sixel.

This adds `TERM_PROGRAM=iTerm.app` → `ProtoSixel` to the auto-detect chain, next to the existing foot/mlterm sixel cases. Images render as real pixels instead of the halfblock mosaic.

The tmux path is intentionally untouched: iTerm2 sets `TERM=xterm-256color`, so `#{client_termname}` can't identify it reliably, and sixel-under-tmux is already documented as conservative-halfblock territory.

### DIFF
--- a/internal/image/capability.go
+++ b/internal/image/capability.go
@@ -107,5 +107,11 @@ func Detect(env Env, cfg string) Protocol {
 	if env.Term == "foot" || env.Term == "mlterm" {
 		return ProtoSixel
 	}
+	// iTerm2: no working kitty graphics implementation (its partial
+	// support fails the startup probe, landing on halfblock), but
+	// solid sixel support — so auto picks sixel for real pixels.
+	if env.TermProgram == "iTerm.app" {
+		return ProtoSixel
+	}
 	return ProtoHalfBlock
 }
--- a/internal/image/capability_test.go
+++ b/internal/image/capability_test.go
@@ -105,6 +105,11 @@ func TestDetect_Sixel(t *testing.T) {
 	cases := []Env{
 		{Term: "foot"},
 		{Term: "mlterm"},
+		// iTerm2 has no working kitty graphics implementation (the
+		// startup probe would time out and downgrade to halfblock),
+		// but it does support sixel — real pixels instead of the
+		// halfblock mosaic.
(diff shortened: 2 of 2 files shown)


gammons:
hi @rfist this looks like a sensible change.  want to make it a non-draft PR?

rfist:
Hi @gammons  — yes, happy to take it out of draft. I'd like to do one more review and test pass on my end first (re-verifying the sixel path against a couple of terminals), so it'll be a little longer before I flip it to ready.
Thanks for taking a look.

j4james:
Is there any reason why you don't just use a standard [device attributes](https://vt100.net/docs/vt510-rm/DA1.html) query to probe for sixel support? An additional benefit from that would be that you could avoid having a timeout in the kitty probe.

Basically you send the kitty query immediately followed by a `DA1` query, and since `DA1` is supported by all terminals, you're guaranteed to receive a response from that. If that response arrives without having received a kitty response, you'll know that kitty is not supported, so you won't need to wait any longer.

rfist:
Superseded by #131, which includes this change verbatim and extends it to WezTerm.

Closing rather than un-drafting because merging the detection change on its own would have
made iTerm2 *worse* than the halfblock it falls back to today. Routing iTerm2 to sixel only
helps if sixel rendering works, and at the time this PR was opened it didn't: images were
re-emitted into the frame string on every render, nothing erased them, and the `v` preview
overlay produced tiled duplicate images that persisted until a forced repaint. #131 fixes
that — images are painted out-of-band against the Bubble Tea frame that actually flushes,
painted once, erased when they move or go away — and then turns detection on for both
terminals.

@gammons — thanks for the nudge back in July, and sorry for the long turnaround. The extra
test pass I mentioned is what surfaced all of the above.

@j4james — your `DA1` suggestion is answered in #131's description rather than here, so it
doesn't get lost when this closes. Short version: you're right, and I'd like to do it as the
next change on its own.

Issue #12, opened 2026-05-15
Feature Request - A way to catch up with unreads (when you have many channels)


kennethito:
Some options that come to mind
* Official slack's unread/catch up feature
* Ability to sort sections by recency similar to official slack (which puts unreads at the top because they are most recent)
* Anything else I'm not thinking of?

Was talking with some coworkers and I think this was the main feature gap.

jpeeler:
One of the first things I always look for is the ability to jump to unread via a shortcut. In slack, it's [two keystrokes](https://slack.com/help/articles/201374536-Slack-keyboard-shortcuts#navigate-conversations): Ctrl+K and then enter. Not ideal, but at least a shortcut exists.

WeeChat's "jump to buffer with activity" [via Alt+A](https://weechat.org/files/doc/weechat/devel/weechat_user.en.html#key_bindings_hotlist) is nicer in that it's just one stroke.

timwis:
Just wanted to share that I'd love to use this, but I very quickly found myself jumping back to the full UI because I rely on the "Activity" view to see the list of messages where I've been mentioned. Now the "Activity" view includes DMs as well, so it's very useful. And the ability to filter to see unread vs read mentions/DMs within that view is quite handy too. (Just to be clear, I don't mean unread channels where I haven't been mentioned—there are far too many of those to ever catch up on! Activity will just show me the ones where I've been notified, e.g. because someone tagged me).

genki-miyachi:
Hey! Took a stab at this and opened a couple of PRs — both runnable locally (`gh pr checkout`, build, point it at your own workspace):

**#109 — Activity view** 🔔 (the @timwis take). A dedicated list in the message pane, mirroring the Threads view, from Slack's internal `activity.feed`: @mentions, thread replies, reactions to your messages, and DMs, with a server-side unread/read filter. Open it from a `🔔 Activity` sidebar row, `:activity`, or a `ctrl+a` toggle.

**#110 — jump-to-unread-channel shortcut** (the @jpeeler / WeeChat Alt+A take). `a`/`A` walk unread channels, marking them read as you go.

Both pass build/vet/test. They touch a few of the same files, so whichever merges first I'll rebase the other. Open on direction, keybindings, or just taking one. 🙂

PR #109, opened 2026-07-27
feat(activity): Activity feed view (mentions, threads, reactions, DMs)

genki-miyachi:
## What

Adds a Slack-native **Activity** view — the desktop client's 🔔 Activity tab — as a dedicated list in the message pane, mirroring the existing Threads view (`⚑ Threads`). It's the "catch up on what pinged me" view a few people asked for in #12.

## How

Sourced from Slack's internal `activity.feed` endpoint, reached with the same `xoxc` browser auth slk already uses for `client.counts`, `subscriptions.thread.getView`, etc. (via `postForm`). It surfaces exactly what the desktop Activity tab does:

| type | shown as |
|------|----------|
| `at_user` / `at_channel` / `at_everyone` / `at_user_group` / `keyword` | mentions |
| `thread_v2` | thread replies |
| `message_reaction` | reactions to your messages |
| `dm` / `bot_dm_bundle` | DMs |

`activity.feed` takes a server-side `unread_only` flag, so the read/unread filter is done by Slack rather than client-side.

## UX

- **Entry:** a `🔔 Activity` sidebar row (below `⚑ Threads`), a `ctrl+a` toggle (the terminal-friendly stand-in for the desktop's <kbd>⌘</kbd><kbd>⇧</kbd><kbd>A</kbd>), and an `:activity` command.
- `u` toggles the unread-only filter (re-fetches).
- `Enter` opens the underlying message/thread in its channel, reusing the existing channel-open / permalink-jump / thread-open paths.
- Rows resolve channel + user names from the same caches the Threads view uses; relative time from `feed_ts`; unread items are bold.

## Design notes

- Mirrors the Threads view at every wiring site (`ViewActivity` alongside `ViewThreads`: view enum, app field/init, msgs, a reducer, the render branch, nav/scroll, the sidebar synthetic row).
- The client flattens each raw item to a small `ActivityItem{Key,Type,IsUnread,ChannelID,TS,ThreadTS,AuthorID,Reaction,…}` so the UI never touches raw JSON.
- **Lenient parsing:** items are decoded individually — a single item with an unexpected field type (Slack's internal endpoints occasionally do this) is skipped, not fatal to the page. A failed fetch returns nil so the existing feed + badge are kept rather than blanked.
- Message-body preview (`messages.list`) and `activity.markRead` are intentionally left as follow-ups; opening an item already marks the underlying conversation read.

## Testing

- `internal/slack/activity_test.go` parses a **sanitized** real `activity.feed` response fixture (all item types) and asserts per-type extraction, plus a regression test that one malformed item is skipped without failing the page.
- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- Driven end-to-end on a live multi-workspace session (fetch → render → `u` filter → `Enter` to open). Easy to try: `gh pr checkout`, build, run against your own workspace.

Happy to rebind any keys (`ctrl+a` / `u`) — those were just what was free. Relates to #12.

_(Full disclosure: built with heavy AI assistance — frontier model, high effort — and reviewed + tested on my end, per the contributing notes.)_


gammons:
hi @genki-miyachi thanks for this!  I love the idea of having the activity feed in the app.  one note when testing, it looks like maybe we're not showing all the data.  
image

if I compare it to the actual activity feed, there's a lot of data about what the activity was, e.g. an emoji added, a reply, etc.
image

genki-miyachi:
@gammons 
Addressed the feedback — rows now show the message body, not just metadata. Each item is a two-line card:

```
@ Alex Rivera  Mention in #announcements         5h ago
   @channel heads up — the release notes are up…
☺ Sam Chen  Reacted in #dev-backend              6h ago
  :tada:  shipped it, thanks!
⚑ Jordan Lee  Thread in #design                  1d ago
  can we get a second pass on the new icons?
```
`activity.feed` only returns refs, so each page now does a second `messages.list` call to hydrate the bodies.

Also rebased onto main to resolve the conflict with the recent notify/usergroups changes — the markup stripping now lives in `slackfmt` and both notify and the activity view share it. Tests added, `build`/`vet`/`test` green, verified live. Let me know what you think of the layout.

gammons:
hi @genki-miyachi thanks for that.  there are a couple of merge conflicts now.  but I think this is ready to get merged in.

genki-miyachi:
@gammons Conflicts resolved — merged latest `main` in.

Only real overlap was the keymap: #110's `a`/`A` stay as-is for unread-channel nav, Activity keeps
`ctrl+a`, and `u` is scoped to the Activity view so they don't collide. Also dropped a gofmt-only
reformat this branch was carrying in `reducer_channels.go`, so that file's diff is down to one line.

`build` / `vet` / `test -race` / lint all green, CI passing, and I re-ran the whole flow by hand.
Ready to merge whenever 🙏

davidnagli:
Hi +1 for this feature, it's the one blocker for me to be able to use this package. I spend 99% of my time in the activity tab and no tool has good keyboard shortcuts so slk will solve a ton of problems for me!

Any progress / blockers on getitng this merged? I'm happy to help if needed

gammons:
@genki-miyachi — apologies, this went stale on my end, not yours. You did two rounds of requested revisions and I told you it was ready, then let it sit. That's on me.

I've now reviewed this side by side with #169, which implements the same feature, and **this is the one I'm merging.** The deciding factor is that this hydrates message bodies via `messages.list` and renders the actual text. #169 renders metadata only (`"direct message"`, `"reply from X"`) — which is precisely the version I sent back in July with those two screenshots. You fixed that; it's the difference between the feature being useful and not.

A few other things I checked and liked:

- 2 requests, lazily on view-open only. No polling, nothing at boot, +60KB binary. Cold start and RSS unaffected.
- `internal/ui/activityview/` mirrors `internal/ui/threadsview/` method-for-method, and `reducer_activity.go` registers like `reducer_threads.go`. Fits the existing shape rather than inventing a new one.
- The 563-line sanitized capture covering 7 item types is the right way to fixture an undocumented endpoint, and you handle `bot_dm_bundle` correctly (payload at `bundle_info.payload.message`, not `item.message`). #169 misses that one and silently drops every bot DM.
- Per-item lenient parsing (`activity.go:157-168`) — one malformed entry is skipped, not fatal. That's the right posture for a private endpoint and I'd like it to be the house pattern.
- The `slackfmt` extraction is motivated by the feature rather than smuggled in, and it fixes a real UTF-8 bug in `notifier.go` on the way (`text[:100]` on a rune boundary).

**Four things and then I merge:**

1. **The conflict is small** — just `keys.go` and `mode_normal.go` against 6647ce4 (the new `d` download binding), caused by your gofmt realignment of the `KeyMap` struct meeting my new field. Re-insert `DownloadFile` and its switch case. Note #171 (landing shortly) reformats the tree and enforces gofmt in CI, so rebase after that and the realignment noise disappears from your diff entirely.

2. **`router.Active()` should be `router.ByID(teamIDStr)`** in both Activity closures in `cmd/slk/main.go`. As written, a workspace switch mid-flight fetches the wrong workspace's feed and labels it with the requested team ID. The reducer's `m.TeamID != a.activeTeamID` guard discards it so nothing corrupts, but it's a latent mislabel and a wasted call. `router.ByID` already exists at `main.go:267`.

3. **`activityNextCursor` is stored but never read** (`app.go:255-258`, `reducer_activity.go:100`). Either wire a "load more" or drop the field — dead state on a view that's otherwise capped at one page.

4. **`case "channel"` reads `item.message`** (`activity.go:189-190`), but that type puts its ref at `bundle_info.payload.channel_entry.latest_message`. Those rows render with an empty `ChannelID` and no-op on Enter. Not in your fixture, so the tests can't see it.

Two follow-ups I'd like but won't hold this for — @agustif got both right in #169 and I'll credit them there:

- The sidebar badge can come free from the `activity_v2` block in the `client.counts` call we already make, which means it's correct at cold start. Right now it reads 0 until you open the view, which is the weakest part of the UX here.
- `ActivityListLoadedMsg` should carry a generation counter so rapid `u` toggling can't land responses out of order.

Thanks for your patience with this one. Ping me when it's rebased.

Closes #12.

PR #110, opened 2026-07-27
feat(sidebar): jump to next/prev unread channel with a/A

genki-miyachi:
## What

Two normal-mode motions for working through unread **channels** without leaving the keyboard:

| key | action |
|-----|--------|
| `a` | open the **next** unread channel |
| `A` | open the **previous** unread channel |

They walk `m.filtered` — the section-sorted set the sidebar actually renders — so they traverse the visible unread dots top-to-bottom, skip the currently-open channel, and wrap around. Muted channels are excluded via the existing `ChannelItem.IsVisiblyUnread` predicate. Because opening a channel already marks it read, repeated presses become a walk-and-clear loop.

## Why

This is the one-keystroke "jump to unread" @jpeeler asked for in #12 (WeeChat's Alt+A, vs. Slack desktop's Ctrl+K + Enter).

To be clear about scope: this is deliberately the **unread-channels** firehose that @timwis noted is often too much to ever catch up on — so I'd treat it as an opt-in companion, not a replacement for the Activity view in #109 (which surfaces only what notified you). Two different takes on #12; happy for either or both.

## Design notes

- New `Sidebar.NextUnread(afterID, dir)` returns the next/prev visibly-unread channel in render order, wrapping, current-channel skipped. Unread channels are never dropped by the staleness filter (`IsStale` returns false whenever `hasUnread`), so `a`/`A` visit exactly what the sidebar shows.
- The handler reuses the existing channel-open path (`sidebar.SelectByID` + `ChannelSelectedMsg`), so focus, mark-read, and collapsed-section expansion behave identically to the fuzzy finder. A status toast shows `No unread channels` when there's nothing else to jump to.
- Keys are hardcoded in `DefaultKeyMap` like every other binding. `a`/`A` because `]`/`[` are already the sidebar-resize keys — happy to rebind.

## Testing

- `internal/ui/sidebar/unread_nav_test.go` (forward/backward, wrap, muted exclusion, current-channel skip, no-other-unread, unknown afterID, no reader).
- `go build ./...`, `go vet ./...`, `go test ./...` pass. Runnable via `gh pr checkout`.

Relates to #12; companion to #109.

_(Full disclosure: built with heavy AI assistance — frontier model, high effort — and reviewed + tested on my end, per the contributing notes.)_

### DIFF
--- a/internal/ui/jump_unread_test.go
+++ b/internal/ui/jump_unread_test.go
@@ -1,0 +1,87 @@
+package ui
+
+import (
+	"strings"
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+	"github.com/gammons/slk/internal/cache"
+	"github.com/gammons/slk/internal/ui/sidebar"
+)
+
+// Pressing `a` from a channel dispatches ChannelSelectedMsg for the next
+// unread channel, carrying its ID/Name/Type (Type drives the message-pane
+// glyph), and moves the sidebar cursor onto it.
+func TestJumpToUnread_EmitsChannelSelected(t *testing.T) {
+	a := NewApp()
+	a.sidebar.SetItems([]sidebar.ChannelItem{
+		{ID: "C1", Name: "one", Type: "channel", Section: "Eng"},
+		{ID: "C2", Name: "two", Type: "dm", Section: "Eng"},
+	})
+	a.SetReadStateReader(func() map[string]cache.ReadState {
+		return map[string]cache.ReadState{"C2": {HasUnread: true}}
+	})
+	a.activeChannelID = "C1"
+
+	cmd := a.handleNormalMode(tea.KeyPressMsg{Code: 'a', Text: "a"})
+	if cmd == nil {
+		t.Fatal("`a` should emit a command when an unread channel exists")
+	}
+	sel, ok := cmd().(ChannelSelectedMsg)
+	if !ok {
+		t.Fatalf("want ChannelSelectedMsg, got %T", cmd())
+	}
+	if sel.ID != "C2" || sel.Name != "two" || sel.Type != "dm" {
+		t.Fatalf("wrong target: %+v", sel)
+	}
+	if it, ok := a.sidebar.SelectedItem(); !ok || it.ID != "C2" {
+		t.Fatalf("sidebar cursor should be on C2, got %+v ok=%v", it, ok)
+	}
+}
+
+// `A` walks backward.
+func TestJumpToPrevUnread_EmitsChannelSelected(t *testing.T) {
+	a := NewApp()
+	a.sidebar.SetItems([]sidebar.ChannelItem{
+		{ID: "C1", Name: "one", Type: "channel", Section: "Eng"},
+		{ID: "C2", Name: "two", Type: "channel", Section: "Eng"},
+		{ID: "C3", Name: "three", Type: "channel", Section: "Eng"},
+	})
+	a.SetReadStateReader(func() map[string]cache.ReadState {
+		return map[string]cache.ReadState{"C1": {HasUnread: true}, "C3": {HasUnread: true}}
+	})
+	a.activeChannelID = "C3"
+
+	cmd := a.handleNormalMode(tea.KeyPressMsg{Code: 'A', Text: "A"})
+	if cmd == nil {
+		t.Fatal("`A` should emit a command")
+	}
+	sel, ok := cmd().(ChannelSelectedMsg)
+	if !ok || sel.ID != "C1" {
+		t.Fatalf("prev unread from C3 want C1, got %+v ok=%v", sel, ok)
+	}
+}
+
+// With nothing else unread, `a` shows a status toast and dispatches no
+// ChannelSelectedMsg.
+func TestJumpToUnread_NoUnread_Toasts(t *testing.T) {
+	a := NewApp()
+	a.sidebar.SetItems([]sidebar.ChannelItem{
+		{ID: "C1", Name: "one", Type: "channel", Section: "Eng"},
+	})
+	a.SetReadStateReader(func() map[string]cache.ReadState { return map[string]cache.ReadState{} })
+	a.activeChannelID = "C1"
+
+	// jumpToUnread returns the toast's clear-tick cmd on the empty path
+	// (not a ChannelSelectedMsg producer); don't invoke it -- it's a 2s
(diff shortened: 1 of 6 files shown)


gammons:
Thanks for this @genki-miyachi — nice feature, and it hooks into the existing selection path cleanly.

I checked it out, read the diff against the surrounding sidebar code, and ran the tests (`go test ./internal/ui/...` all green, `go vet ./...` clean). A few things to fix before merge.

## What's good

- `SelectByID` + `ChannelSelectedMsg` is the right pattern — it matches the channel finder path, so `reduceChannelSelected` does the real work (activate, retarget, focus, cache tier, `MarkRead`). No "cursor moved but conversation never loaded" bug.
- Viewport scroll-to-reveal works, for a non-obvious reason: `SelectByID` mutates `m.cursor`, which trips the `snappedSelection != cursor` check in `View()`. Easy to miss; you got it right.
- Reusing `item.IsVisiblyUnread(state[item.ID])` instead of re-deriving `HasUnread && !IsMuted` keeps the single-source-of-truth contract, and mute-exclusion falls out for free.
- Walking `m.filtered` rather than `m.items` is the correct choice — that's the section-sorted render order.
- `NextUnread` as a pure query on the sidebar model with the side effects staying in the App layer is good separation.

## Needs fixing

### 1. Anchor off-by-one — first press lands on the wrong end (`internal/ui/sidebar/model.go:355-366`)

The comment says *"when it isn't in the visible set, start from the top (forward) or bottom (backward) so the first press still lands"* — but the code does the opposite. `start = 0` combined with a loop starting at `off = 1` means position 0 is checked **last**, after a full wrap around. Mirrored for backward.

Confirmed with a scratch test (C1/C2/C3, C1 and C3 unread, `afterID = ""`):

```
forward  from unknown anchor: id="C3" ok=true   (want C1)
backward from unknown anchor: id="C1" ok=true   (want C3)
```

This won't show up in manual testing because a channel is almost always already open, so the anchor is found. It bites on a fresh start before opening anything, and after any `SetItems` rebuild where `activeChannelID` drops out of `filtered`. `TestNextUnread_UnknownAfterID` misses it only because C1 happens to be *read* in that fixture.

Fix is to move the sentinel one step behind the first real position:

```go
start := -1
if step < 0 {
    start = n
}
```

The existing `((start+step*off)%n+n)%n` normalization already handles `-1` and `n` correctly (`start=-1, off=1 → 0`; `start=n, off=1 → n-1`), and `off` in `1..n` still visits every position exactly once. Worth a regression test with the first visible row unread.

### 2. The doc comment on `NextUnread` isn't accurate (`internal/ui/sidebar/model.go:334-339`)

> *"a/A visit exactly what the sidebar shows."*

`m.filtered` is built by `rebuildFilter`, which doesn't know about collapse state — collapse is applied later in `rebuildNav` and `buildCache`. So `filtered` includes channels in collapsed sections, which aren't shown. Since `New()` collapses `Channels` and `Apps` by default, pressing `a` on a typical workspace will jump into the firehose section and permanently expand it (`SelectByID`'s slow path calls `ToggleCollapse`).

To be clear, that auto-expand is pre-existing `SelectByID` behavior, not something this PR broke, and ignoring collapse for an unread skim is defensible — Slack's own does. But right now it's accidental rather than chosen, and undocumented. Either keep it and have the comment say so explicitly (plus a test pinning the expand side effect), or skip collapsed sections in the loop. Your call which, I lean toward keeping it.

(The staleness-filter claim in the same comment does check out.)

### 3. `wiki/Keybindings.md` not updated

`U` is documented at line 34 and the README points people there as "every key, every mode." Something like:

```
| `a` / `A` | Normal | Jump to next / previous unread channel (wraps) |
```

### 4. No App-level test for `jumpToUnread`

The sidebar query is covered well, but nothing asserts that `a` emits `ChannelSelectedMsg` with the right ID/Name/Type, that the cursor actually moved, or that the empty case toasts instead of selecting. There's precedent in `app_test.go` (`TestMarkUnreadOfSelected_*`) and `open_links_test.go`.

## Minor

- **Toast wording** (`mode_normal.go:327`): when the open channel is the only unread one, the user sees "No unread channels" while looking at an unread channel. "No other unread channels" is accurate in every case.
- **Test coverage gaps**: no unreads at all (distinct from `NoOtherUnread`, which has one), DM / `group_dm` / `app` types (`Type` is forwarded straight into `ChannelSelectedMsg` and drives `messagepane.SetChannelType`), and a non-empty `m.filter`.
- **Placement**: sibling handlers (`markUnreadOfSelected`, `openLinksOfSelected`, `copyPermalinkOfSelected`) all live in `app.go`; `mode_normal.go` is otherwise pure dispatch. Not worth churning on its own.

---

Design is sound and there's nothing architectural to redo here — (1) is a four-line change plus a test, and (2) just needs a decision. Happy to take another look once those land.

genki-miyachi:
@gammons 

Thanks for the super thorough review — all fixed in the follow-up commit (`4ef556f`), squashable on merge.

**1. Anchor off-by-one** — nice catch, that was a real bug. Moved the sentinel one step behind the first position (`start := -1` forward / `n` backward); the found-anchor path is unchanged. Added `TestNextUnread_UnknownAnchor_LandsOnNearestEnd` (your C1/C3-unread scratch case) — confirmed it fails on the old `start := 0` and passes now.

**2. Collapse behavior** — went with keeping the auto-expand, as you leaned. Rewrote the `NextUnread` doc comment to say so explicitly (walk includes unread channels in collapsed sections; `SelectByID` expands on jump; matches the desktop client's "all unreads" skim), and added `TestNextUnread_ReachesCollapsedSectionAndExpands` pinning both the reach and the expand side effect.

**3. `wiki/Keybindings.md`** — added the `a` / `A` row (next to `n`/`N`, since both are next/prev-with-wrap).

**4. App-level test** — added `jump_unread_test.go`: asserts `a`/`A` emit `ChannelSelectedMsg` with the right ID/Name/Type, the cursor actually moves, and the empty case toasts instead of selecting.

**Minor** — toast now reads "No other unread channels"; added coverage for no-unreads-at-all, DM/`group_dm` type forwarding, and a non-empty filter. Left `jumpToUnread` in `mode_normal.go` per your "not worth churning" note — happy to move it to `app.go` if you'd prefer.

`go build`/`vet`/`test` green, and I re-ran it end-to-end. Thanks again!

gammons:
lgtm!  thank you @genki-miyachi

Issue #111, opened 2026-07-27
Failed to mint token: mint token: status 403


larsks:
slk correctly identifies the workspaces to which I am authenticated in the desktop app, but it fails immediately with an error:

```
Failed to mint token: mint token: status 403
```

I didn't see any other reports of this error here or via a google search. Is this something I can correct?

I'm using slk:

```
slk 0.12.0 (commit 8149c3b18ed04c259efe5feb545d040ab043d922, built 2026-07-27T14:32:15Z)
Unofficial Slack client. Not affiliated with Slack Technologies, LLC.
Uses Slack's internal browser protocol; may violate Slack's TOS. Use at your own risk.
```

The desktop version is slack (`slack-4.49.89-0.1.el8.x86_64`) on Fedora 44.

gammons:
Hi @larsks thanks for the report. 

Some context first: slk very recently changed how it authenticates (as of v0.12.0). It used to have you paste a token out of the browser DevTools.  It now reads your existing Slack desktop app session automatically. That change was to fix #5 (browser-token extraction was tripping enterprise security and signing people out). So the code you're hitting is brand new.

Are you on Slack Enterprise Grid? (looks like you're at Red Hat so I am assuming so) Enterprise Grid behind a corporate proxy is exactly the environment where this request can get blocked, and it's closely related to #5.

As for the likely cause: to scrape that token, slk was requesting https://<workspace>.slack.com but sending the headers of an in-app API/XHR call (Origin: https://app.slack.com, Sec-Fetch-Mode: cors) instead of the headers a browser sends for a normal page navigation. That mismatched signature is exactly what a strict corporate proxy or TLS-inspecting gateway tends to reject with a 403 — it slips through on most networks, which is why it wasn't caught earlier.

I've pushed a candidate fix (#114) that makes that request look like a real page navigation. Could you try it?

```
go install github.com/gammons/slk/cmd/slk@fix/111-mint-403-navigation-headers
slk --add-workspace
(or clone, git checkout fix/111-mint-403-navigation-headers, go build ./cmd/slk)
```

If that fixes it, I'll get it merged. If you still get a 403, it likely means your network path is blocking non-browser access to the workspace page outright — in which case a few details would help:
- Enterprise Grid workspace? (probably yes)
- Corporate VPN / proxy / TLS inspection (Zscaler and friends)?
- Does the Slack web app (in a browser) work on the same machine/network?

Thanks for helping track this down!

larsks:
[...]

PR #114, opened 2026-07-28
fix(slack): use navigation-shaped headers for token minting (#111)

gammons:
Loading the workspace page to scrape api_token is a top-level navigation, but it went through the XHR BrowserTransport, which stamps it with Origin: https://app.slack.com and Sec-Fetch-Mode: cors. That contradictory signature (a page load wearing fetch/CORS headers) is rejected with HTTP 403 by strict corporate edges/proxies, so onboarding failed with 'mint token: status 403'.

Mint with a plain client and navigation-appropriate headers (Sec-Fetch-Mode: navigate, Sec-Fetch-Dest: document, no Origin, Accept: text/html) instead, matching what a browser sends on a page load (and what gh-slack does). Using a plain client with no cookie jar also stops sending the d cookie twice. The ongoing API/WebSocket traffic keeps the browser-header treatment (#5).

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -492,6 +492,12 @@ func main() {
 				os.Exit(1)
 			}
 			os.Exit(0)
+		case "--dump-mint":
+			if err := dumpMint(); err != nil {
+				fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+				os.Exit(1)
+			}
+			os.Exit(0)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


gammons:
Superseding this approach. The `--dump-mint` diagnostic (from this branch) revealed the real root cause: on Enterprise Grid, loading the workspace page redirects to the enterprise host and returns a sign-in page, so scraping `api_token` from a page load can't work for Grid at all. Instead of minting via a network page load, we'll read the xoxc token directly from the desktop app's own `localConfig_v2` store (Local Storage), which contains the token for every team including Grid. Closing in favor of that work; the `--dump-mint` diagnostic will carry over.

PR #120, opened 2026-07-29
Fix/kde secret service

fleeco:
Here is the breakdown shamelessly taken from Claude

Why slk couldn't connect: two independent halves of one credential

A Slack browser/desktop session is two secrets that must be presented together:

| Secret | What it is | Where it's stored | Encrypted? |
|---|---|---|---|
| `d` cookie (`xoxd-…`) | the session auth cookie | Cookies SQLite DB | **Yes** — AES, key in OS keyring |
| xoxc token (`api_token`) | the per-workspace API token | Local Storage LevelDB | No — plaintext JSON |

slk needs both. An xoxc token is explicitly worthless without its paired d cookie — that's the whole design of "cookie tokens." The two fixes each unblock one of these secrets, on different failure modes, and both are required.

---
#1 — Read the token from LevelDB instead of scraping api_token from the page

The old way (dead). slk minted the xoxc token by doing an HTTP GET of https://<workspace>.slack.com and regex-scraping "api_token":"xoxc-…" out of the returned HTML (internal/slack/mint.go). For years, Slack embedded the token in that page's boot payload.

Why it broke. Slack's modern web client (client-v2) no longer inlines the token. Our diagnostics caught it red-handed: the page came back 200 OK, 156 KB, data-app="client-v2", and zero occurrences of xoxc- or api_token anywhere in it. The token simply isn't on the page anymore — the JS app boots and gets it from local storage instead. So every mint returned "token not found," slk fell back to a stale cached token, and the workspace never connected (empty sidebar, no WebSocket).

The fix. Read the token from the same place the real client keeps it — the desktop app's Local Storage LevelDB (~/.config/Slack/Local Storage/leveldb), under the localConfig_v2 key → .teams.<TEAM_ID>.token. New code: internal/slackdesktop/Tokens(). Three details that make it reliable:

1. Copy-to-temp. LevelDB is single-writer and Slack holds it open, so we copy the DB dir and read the copy read-only — the same trick already used for the Cookies DB. That's why you didn't have to quit Slack.
2. Key matching. Chromium stores values under origin-prefixed keys like _https://app.slack.com\x00\x01localConfig_v2; we iterate and match any key ending in localConfig_v2 rather than guessing the exact origin.
3. Value decoding. Chromium tags each value with a leading byte (0x00=UTF-16LE, 0x01=Latin-1); we strip and decode it, then JSON-parse.

Onboarding and the per-launch refresh now ask LevelDB first and only fall back to the old scrape-mint for legacy workspaces that still inline the token.

---
#2 — The KDE secret fix, and whether it's still needed

What it fixes. To decrypt the d cookie, slk needs the "Slack Safe Storage" AES key, which Chromium/Electron stores in the OS Secret Service. The lookup lives in keyringPassword() (internal/slackdesktop/password_linux.go). Before this branch, slk only queried the Chromium libsecret schemas:

xdg:schema = chrome_libsecret_os_crypt_password_v2   (and _v1)

On a KDE desktop, Slack doesn't use those. KDE routes the Secret Service through KDE Wallet via QtKeychain, which stores the key under a completely different schema:

xdg:schema = org.qt.keychain,  server = "Slack Keys",  user = "Slack Safe Storage"

So on KDE the old code found nothing → ErrNoSecretService/ErrKeyringLocked → couldn't decrypt the d cookie at all. The fix (32e051a) refactors keyringPassword to try a list of schemas — the two Chromium ones and the QtKeychain one — returning the first unlocked match. That's the entire change: it teaches slk where KDE hides the key.

Was it still necessary after the LevelDB change? Yes — they're orthogonal, and the logs prove it. The LevelDB fix gets the token; the KDE fix gets the key to decrypt the cookie. Recall this line from the debug logs:

d cookie len=239 xoxd_prefix=true

That xoxd_prefix=true means the d cookie decrypted correctly — which could only happen because keyringPassword() found the AES key. On your KDE session that key is in KDE Wallet under org.qt.keychain, which only the fixed code queries. Strip the KDE fix out and that decrypt fails, you get a garbage cookie, and Slack rejects every request no matter how good the token is.

Note the neat separation: the token path (LevelDB) needs no keyring because localConfig_v2 is plaintext; only the cookie path needs the KDE fix. Two secrets, two storage mechanisms, two fixes.


gammons:
thanks @fleeco !

PR #121, opened 2026-07-30
Enterprise Grid parity: request-shape, call-pattern and fetch-scope (phase 1 of 3)

gammons:
Umbrella PR for Enterprise Grid parity (#5, #111). **Phase 1 of 3 is complete; Phases 2 and 3 will be pushed to this same branch.** Draft until all three land.

## Why

Enterprise Grid orgs sign slk users out and email them a security notice naming *"data scraping, excessive file downloads, connecting to Slack from a Tor exit node, or using a third-party client."* Three prior attempts didn't fix it — browser-like headers (#22), network token minting (#106), `localConfig_v2` token reads (#115). In #111 a tester was signed out **twice** while using the desktop app's own token and cookie, which ruled out credentials.

So I captured the **official Slack web client's** traffic — 8 HAR captures on a busy workspace (cold boot, warm boot, two channel switches, scrollback, quick switcher, channel search, and a 90-second disconnect/reconnect) — and diffed it against what slk does.

**This disconfirmed my own working theory.** I assumed slk was downloading too many assets. The official client pulls **68 MB over 337 requests at 80-way concurrency** at boot; slk caps at 4 and fetches thumbnails only. On assets slk is already an order of magnitude politer than the real client.

The actual divergence is three separate layers:

| Layer | Problem | Status |
|---|---|---|
| **1. Request shape** | No `_x_*` telemetry params, no `sec-ch-ua` client hints, `Chrome/120` UA (30 versions stale), and a `Referer` the real client never sends. Separable by a single log predicate, before anyone looks at behaviour. | **This PR** |
| **2. Call pattern** | slk *enumerates* — `users.list` (~50 pages), `conversations.list` (every public channel), `conversations.history` for **every channel ever visited**, plus `conversations.replies` per thread. The official client never enumerates in any capture; it revalidates a local cache conditionally. This is the "data scraping" signature. | Planned |
| **3. Fetch scope** | Assets are fetched for the whole 50-message buffer instead of the viewport, head-of-line-blocking behind a 4-wide semaphore. Cause of the slowdown on busy workspaces, **not** of the sign-outs. | Planned |

Design: `docs/superpowers/specs/2026-07-30-enterprise-grid-bootstrap-design.md`

## What Phase 1 does

All changes are in `internal/slackhttp`, which already sat beneath both API call paths (slack-go and the hand-rolled `postForm`), plus the wiring in `internal/slack`.

- Chrome 120 → 150, with the UA and `sec-ch-ua` client hints derived from one constant so they cannot drift
- Adds the three client hints plus `cache-control` / `pragma` / `priority`; removes the `Referer` (0 of 279 captured API requests carry one)
- Query params and body fields emitted **in the order the real client emits them** — `url.Values.Encode()` sorts, and 0 of 163 captured requests carried alphabetized params
- Per-destination header sets: XHR, WebSocket upgrade, and image fetches each get their own, because Chrome sends three different sets
- `_x_version_ts` sourced from `client.shouldReload`, persisted per workspace, seeded and refreshed on connect

## Bugs found that weren't in the plan

**The WebSocket upgrade was wrong, and had been.** `client.go` asserted `Sec-Fetch-Dest: websocket` was "the value a real browser sends." Chrome sends **no `Sec-Fetch-*` at all** on an upgrade, plus no `Accept`, no `sec-ch-ua`, no `Priority`. slk was sending 4 headers Chrome omits; consolidating header sources briefly added 4 more. A long-lived socket with a header set no Chrome emits is a stable signature. Now has its own verified 5-header set.

**Asset fetches got worse before they got better.** Removing the `Referer` globally also removed it from image requests — where Chrome *does* send one. On the 337-requests-per-boot path. Fixed by giving images their own header set (`sec-fetch-dest: image`, `mode: no-cors`, image `Accept`, `Referer` present, `Origin` absent).

**`_x_reason` was emitted on exactly one call site.** The context carrier was built; no caller was taught to use it. That left `body has _x_mode AND NOT _x_reason` matching ~100% of slk traffic and ~6% of the client's — replacing a generic-HTTP-client signature with a Slack-client-shaped-but-wrong one, which is worse. Now defaults per endpoint from nine observed values.

**Three endpoints were making live outbound calls during tests.** `GetUnreadCounts`, `callChannelSectionsList` and `GetStarredChannels` each built their own HTTP client, so they bypassed the test harness and could silently lose the envelope. Mutating them returned real `not_authed` responses from Slack's edge. Now routed through the shared client; the suite passes under `unshare -rn` with loopback only.

**`config.toml` gained concurrent writers.** The background `version_ts` refresh fires once per workspace, all connecting in parallel. A concurrency test failed **14/20 runs** with lost updates and a truncated-read-persisted-back case. Fixed with a mutex plus atomic temp-file-and-rename across all four savers (the pre-existing theme/width savers needed it too).

**`_x_id` is not unique in the real client** — 163 values across the captures, 160 distinct. The plan asserted uniqueness and a clamp was written to satisfy it; forcing uniqueness would give slk a perfectly monotonic sequence where Chrome shows collisions. Clamp removed.

## Testing

Every behavioural claim is pinned to the captures, and the fixtures drive the tests rather than the reverse — `golden_test.go` hardcodes zero param names and reads `testdata/official-request-shape.json`.

The raw HARs are **not** committed (live `xoxc` tokens and message content). `testdata/capture-evidence.json` is a redacted digest of the measured aggregates, added after three reviewers in a row couldn't verify the numbers — one filed a false-positive blocker as a result.

Each task was mutation-tested. Several mutations survived on first attempt and are now caught, including two vacuous tests where `captureRT` shared a `Body` the inner transport had already drained.

`go build`, `go vet`, `go test ./... -race`, `golangci-lint run ./...` all clean; 50 packages, 0 failures.

## Known residual divergences

Recorded in full in `docs/superpowers/plans/2026-07-30-grid-parity-phase1-outcomes.md`:

- **Body encoding** — slk sends `x-www-form-urlencoded`; all 163 captured bodies are `multipart/form-data`. Deferred; it's a transport-wide rewrite.
- **Body param order** — business params are alphabetical via `Encode()`. Deliberately *not* half-fixed: slack-go builds ~50 endpoint bodies the same way and is out of reach, so ordering only the ~6 hand-rolled ones would give slk two distinguishable shapes instead of one.
- **`Accept-Encoding: gzip`** — Go's stdlib default; Chrome sends `gzip, deflate, br, zstd`. Fixing it means manual decompression.
- **`Authorization: Bearer` on image fetches** — the real client never sends that header anywhere. Highest-value follow-up, but `main.go:687` documents that the `d` cookie alone returns a login page, and the HAR can't settle it (Chrome stripped cookies from the export). Needs its own scoped investigation.

## What this does not do

**It probably does not fix the sign-outs on its own.** Layer 2 is the scraping signature, and per the reconnect capture slk replays its full fan-out on every sleep, wifi change and VPN flap — several times a day, not once at boot.

If detection is TLS/JA3-level (raised by @Icantjuddle in #5), none of this helps. Phase 1 being small and independently shippable is what makes it a cheap read on that.

**Not asking anyone to test yet.** @larsks was signed out of his work org twice helping diagnose #111, and @raff and others took the hit on #5. I'll say explicitly when all three phases are ready, and nobody should feel obliged.


gammons:
## Status at ready-for-review

All three phases are in. Measured on a two-workspace setup (105 and 39 channels),
`SLK_DEBUG=1`, shutdown tally:

| endpoint | before | after |
|---|---|---|
| `conversations.history` | 144 | 1 |
| `subscriptions.thread.getView` | 79 | 0 |
| `conversations.list` | 4 | 0 |
| `users.list` | 2 | 0 |
| session total | 270 | 37 |

Cold cache (a fresh install — the path every new user takes) went from
**37,573 `users.info` requests to 235**, and from an unresponsive UI at load
average 30 to no measurable spike. That regression was found by the QA
checklist, not by the test suite.

`users.list`, `conversations.list` and the reconnect fan-out are gone from the
interfaces that could reach them, so they fail a test rather than a review if
reintroduced.

### Not implemented

- **Mention picker still filters the local roster** rather than calling
  `edge.UsersSearch` (plan Task 11b). Removing the per-member resolution
  shrank its candidate list to "users slk has seen" — a deliberate trade, but
  a real reduction until 11b lands.
- **`conversations.view`'s history is fetched and discarded.** The opened
  channel still renders through the old cache + `GetHistory` path. Its `users`
  and `emojis` payloads *are* used; the messages are not, and the
  `[]slack.Message` → `[]json.RawMessage` conversion has never been validated
  against a real render.
- **Success criterion 1 (boot ≤ 10 API calls) is unmet** at ~18 per workspace.
  Criteria 3, 4 and 5 are met; criterion 2 (O(1) reconnect) is met in tests
  only. Remaining duplicates are itemised in the Phase 2b outcomes doc —
  two of them (`usergroups.list`, `stars.list`) are blocked on capture
  evidence rather than effort, because both existing captures show empty
  arrays and guessing the shape is how this project has been burned twice.

### Not verified

Manual QA (`docs/superpowers/plans/2026-08-03-grid-parity-qa-checklist.md`) is
through section 4 of 11. Sections 5-10 — mention picker, a real 90-second
outage and reconnect, Threads, finder, general regressions — are outstanding.
The reconnect path is the one I would watch: it is new code that runs on every
wifi flap and it has never seen a real outage.

**Nobody has tested any of this against Enterprise Grid**, which is the problem
it exists to solve. Two contributors have already been signed out helping
diagnose the original issue. The `conversations.view` `channel` parameter slk
now depends on — with a fallback — has never been exercised there.

PR #128, opened 2026-08-05
fix(slackdesktop): detect flatpak and snap Slack config dirs on Linux

### LINKED ISSUE #123: Slack desktop app flatpak version
With the new update, is it possible to sign in using the slack flatpak? 

```
slk -- Add Workspace

Reading your signed-in workspaces from the Slack desktop app.

  Slack desktop app not found. Install it and sign in, then retry.
Error: slack desktop app config directory not found
```

If not, is there any workaround to use the old browser method?

gammons:
## Summary

PR #106's desktop-app auth flow only looked at `$XDG_CONFIG_DIR/Slack` or
`~/.config/Slack` on Linux, so users of the **Slack flatpak** (reported in
#123) and **snap** got `slack desktop app config directory not found` even
when signed in to the desktop app.

`configDirForOS` now probes candidate locations in order and picks the first
that exists:

1. `$XDG_CONFIG_HOME/Slack`
2. `$XDG_CONFIG_DIR/Slack` (legacy, kept for back-compat)
3. `~/.config/Slack` (native packages)
4. `~/.var/app/com.slack.Slack/config/Slack` (**flatpak**)
5. `~/snap/slack/current/.config/Slack` (**snap**)

Native installs are preferred over packaged ones when both exist. When
nothing exists, the error still references the conventional path.

**Cookie decryption needs no changes:** the Slack flatpak stores its
"Slack Safe Storage" secret on the host secret service using the same
chromium libsecret schema (`application=Slack`), and the existing v10
"peanuts" fallback covers setups without a reachable secret service.

Re: #123's ask for the old browser-token flow as a workaround — that was
removed intentionally in #106 because it triggers Enterprise Grid anomaly
detection (#5); fixing config-dir detection is the right resolution.

Fixes #123

## Test plan

- [x] New table-driven tests: flatpak-only, snap-only, native-over-flatpak
      precedence, flatpak-over-snap precedence, `XDG_CONFIG_HOME` preference,
      XDG-set-but-missing fall-through (written test-first, verified failing)
- [x] `go test ./... -race` green, `go vet ./...` clean
- [ ] End-to-end on a real flatpak install — would appreciate a confirmation
      from the #123 reporter

### DIFF
--- a/cmd/slk/onboarding.go
+++ b/cmd/slk/onboarding.go
@@ -132,6 +132,8 @@ func desktopErrorMessage(err error) string {
 		return "Your system keyring is locked. Unlock it (log in to your desktop session) and retry."
 	case errors.Is(err, slackdesktop.ErrNoSecretService):
 		return "No system keyring/secret service found. slk needs it to read the Slack session."
+	case errors.Is(err, slackdesktop.ErrSecretNotFound):
+		return "No Slack entry found in your keyring. Sign in to the Slack desktop app (and make sure it uses the system keyring), then retry."
 	case errors.Is(err, slackdesktop.ErrDecryptFailed):
 		return "Could not decrypt the Slack session cookie. Please file an issue with your OS + Slack version."
 	default:
--- a/internal/slackdesktop/configdir.go
+++ b/internal/slackdesktop/configdir.go
@@ -21,10 +21,29 @@ func configDirForOS(goos string, getenv func(string) string, exists func(string)
 		}
 		return second
 	default: // linux and others
-		if x := getenv("XDG_CONFIG_DIR"); x != "" {
-			return filepath.Join(x, "Slack")
+		home := getenv("HOME")
+		var candidates []string
+		if x := getenv("XDG_CONFIG_HOME"); x != "" {
+			candidates = append(candidates, filepath.Join(x, "Slack"))
 		}
-		return filepath.Join(getenv("HOME"), ".config", "Slack")
+		if x := getenv("XDG_CONFIG_DIR"); x != "" {
+			candidates = append(candidates, filepath.Join(x, "Slack"))
+		}
+		candidates = append(candidates,
+			filepath.Join(home, ".config", "Slack"),
+			// flatpak (com.slack.Slack)
+			filepath.Join(home, ".var", "app", "com.slack.Slack", "config", "Slack"),
+			// snap
+			filepath.Join(home, "snap", "slack", "current", ".config", "Slack"),
+		)
+		for _, c := range candidates {
+			if exists(c) {
+				return c
+			}
+		}
+		// Nothing found: fall back to the primary location so the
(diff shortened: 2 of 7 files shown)


MrSmoke:
Im getting the following error message

```
  No system keyring/secret service found. slk needs it to read the Slack session.
Error: no system secret service available
```

gnome-keyring is installed and running. 

I think the keyring logic will also need updating to handle flatpak/snap as well

gammons:
@MrSmoke Progress — the flatpak config dir is found now; the failure moved to the keyring lookup. I just pushed an update that splits the conflated error, so a rebuild will tell us exactly which side is failing:

- **"No system keyring/secret service found"** → slk can't reach the Secret Service D-Bus API at all
- **"No Slack entry found in your keyring"** → the service is reachable but has no Slack item (e.g. the flatpak stored it under different attributes)

Could you rebuild from this branch and also paste the output of:

```bash
secret-tool search application Slack
busctl --user list | grep -i secret
```

The first one shows whether a 'Slack Safe Storage' entry exists in gnome-keyring and what attributes it has — that tells us what to search for.

gammons:
Picking this back up — it's finished, not abandoned, and it should land.

Re-verified against current main:

- The bug is still live. `internal/slackdesktop/configdir.go:23-28` has no flatpak or snap handling, and `git log 24e4df3a..origin/main -- internal/slackdesktop/configdir.go` is empty, so main has gained nothing overlapping since this branch forked.
- It rebases onto main with **zero conflicts**, despite main having changed both `password_linux.go` (72e4059) and `cmd/slk/onboarding.go` (7cad2d6) since. Post-rebase: build, vet, and `go test ./internal/slackdesktop/... ./cmd/slk/... -race` all pass.
- It also quietly fixes a real bug I'd forgotten about: main only reads the non-standard `XDG_CONFIG_DIR` and never the spec's `XDG_CONFIG_HOME`. This adds the correct one while keeping the legacy name as a fallback candidate.

**The open question that stalled this is now answered.** I'd asked @MrSmoke for `secret-tool search application Slack` output after a keyring failure and never heard back, so the "end-to-end on a real flatpak install" box stayed unchecked. @yml has since confirmed it independently in #158 — built this branch, ran it against flatpak Slack on Manjaro, all 8 workspaces connected. That also closes the open risk that the host secret service might not serve the flatpak's Safe Storage key. It does.

@MrSmoke's keyring failure looks environment-specific and is a separate concern from config-dir detection; I'll split it into its own issue rather than let it block this.

Plan: rebase onto main (after #171 lands, which fixes a flaky test and enforces gofmt), let CI re-run, merge. Note for sequencing — **do not squash-merge this** while #158 is stacked on it; #158's branch is based on this one, so a squash would rewrite these commits and strand it. Rebase-merge or merge commit.

Fixes #123.

MrSmoke:
Sorry I've been on holidays. I can verify the fix next week

PR #131, opened 2026-08-06
Inline images on sixel terminals (WezTerm, iTerm2, foot, mlterm)

rfist:
# Inline images on sixel terminals (WezTerm, iTerm2, foot, mlterm)

## What this fixes

Inline images render as a blurry half-block mosaic on WezTerm and iTerm2 today. This makes them render as real images.

The cause isn't that those terminals lack image support — both speak kitty *graphics*. It's that slk's kitty renderer positions images using the **unicode placeholder extension** (`U=1` with U+10EEEE characters written into the text, `internal/image/kitty.go`). That's a kitty-specific extension, not part of the kitty graphics protocol. A terminal can fully support kitty graphics and still be unusable on that path — WezTerm is exactly that case, and routing it to kitty makes it print the placeholder codepoints as literal glyphs. So sixel is the only viable route for WezTerm, iTerm2, foot and mlterm.

## Why this is more than "emit sixel bytes"

Kitty gets three things from the terminal that sixel doesn't: cell-aware placement, image identity (so the terminal repaints them), and clipping. Sixel pixels are fire-and-forget — nothing repaints them, nothing erases them, and they ignore pane borders entirely. slk has to own the screen region itself.

The previous approach appended sixel bytes into the frame string. That forces re-emission on every render (~283 KB per frame per visible image, measured) because Bubble Tea rewrites a line whenever its content changes — and it offers no way to clear a region an image has left behind.

Instead, `App.View()` publishes an immutable snapshot of the placements it wants into a frame store, tagged with an ID. A writer wrapping the program output (`internal/image/sixelframe.go`) takes the exact frame that actually flushed and reconciles it against what's on screen: unchanged → zero bytes, moved or gone → erase, new or changed → paint.

Three details that took real-terminal iteration to get right:

- **Erase and paint land on opposite sides of the text diff.** ECH blanks the cells it clears, so an erase running after the diff wipes text the diff just wrote — the symptom was messages disappearing while scrolling. Erases now go immediately after the synchronized-output start, paints immediately before its reset.
- **The painter is transactional.** `Plan` computes bytes and next state without mutating; `Commit` runs only after a full successful write. A short write can't leave the painter believing an image is on screen that isn't.
- **Content identity is a hash, not a byte length.** Two different images with the same dimensions and palette depth encode to the same length, so a length-keyed placement treated a slot swap as unchanged and never repainted.

The preview overlay also memoizes its encoded image. `View()` runs after every bubbletea message and the compositor deliberately never memoizes overlay frames, so without it the overlay re-encoded on every keystroke and mouse motion — measured at ~1.4s per encode at full-pane size, which made opening and closing the preview visibly lag. Repeat renders are now ~44µs.

## Scope and limits

- The thread pane and avatars still render half-block on sixel. Avatars are a natural follow-up now that emission is change-gated (the original rationale — re-emitting per avatar per redraw would dominate bandwidth — no longer holds).
- Sixel under tmux is unchanged and still conservative: no probe-verify, and passthrough is bandwidth-hostile. WezTerm-under-tmux therefore still takes the old path.
- **All output now flows through the wrapping writer, not just sixel.** For every other protocol it's a serialized pass-through: frames carry no marker, so no plan is computed and the bytes are forwarded unchanged. `Fd`/`Read`/`Close` delegate so Bubble Tea's `term.File` assertion still finds the TTY. Flagged explicitly because it's the part of this PR that touches users who don't use sixel at all.

## On detection (and #100)

This supersedes #100, which carried the `iTerm.app` → sixel detection change on its own. That change is included here verbatim and extended to WezTerm. Merging it standalone would have made things *worse* for iTerm2 users than the halfblock they have today: it routes them onto sixel, and until the painting work in this PR, sixel meant tiled duplicate images in the preview overlay and stale pixels that survived until a forced repaint. That's why it stayed a draft.

@j4james — on #100 you asked why detection doesn't use a `DA1` device-attributes query instead of matching `TERM_PROGRAM`, and pointed out that pairing it with the kitty query gives a guaranteed response to synchronize against, removing the need for a timeout. You're right, and it would replace the weakest part of this PR: the terminal-name matching in `Detect`, plus the 200ms probe timeout that is exactly what made WezTerm silently fall back to halfblock in the first place.

I've kept this PR to the rendering work, which is independent of *how* the protocol is chosen, so the detection rework can land as its own reviewable change rather than riding along with 2.8k lines of painting logic. Happy to take on that as the immediate follow-up.

## Testing

Sixel positioning can't be verified in CI — there's no terminal. Unit coverage pins what can be pinned: frame correlation and pruning, erase/paint ordering around the sync window, marker stripping, title forwarding, short-write and failed-write non-commit, and an end-to-end test driving a real Bubble Tea program through the writer (`TestSixelFrameOutput_BubbleTeaFlushContract`).

Manually verified in **iTerm2**, **WezTerm** and **Ghostty** — inline images, the `v` preview overlay, scrolling past images, and selecting rows containing images. Ghostty is on the kitty path and was checked specifically because of the output-path change noted above; it is unaffected.

`go build`, `go vet` and `go test ./...` pass.

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -899,6 +899,16 @@ func run() error {
 
 	// Create app
 	app := ui.NewApp()
+	// Frame-correlated sixel output: the App publishes immutable
+	// placement snapshots into sixelFrames; terminalOutput strips the
+	// internal frame marker from Bubble Tea's window title, takes the
+	// exact flushed frame, and paints its sixel operations after the
+	// text diff. Kitty uploads share the same serialized writer so
+	// byte streams from different goroutines can't interleave.
+	sixelFrames := imgpkg.NewSixelFrameStore()
+	terminalOutput := imgpkg.NewFrameOutput(os.Stdout, sixelFrames)
+	imgpkg.KittyOutput = terminalOutput.SideChannel()
+	app.SetSixelFrameStore(sixelFrames)
 	app.SetHelpFooter(versionpkg.ModalFooter(version))
 	app.SetClipboardAvailable(clipboardOK)
 	if sr := notify.NewStatusReporter(cfg.Notifications.StatusCommand); sr != nil {
@@ -1015,6 +1025,12 @@ func run() error {
 	// Cell pixel metrics for sizing decisions.
 	pxW, pxH := imgpkg.CellPixels(int(os.Stdout.Fd()))
 	debuglog.ImgRender("cell pixels: %dx%d", pxW, pxH)
+	// Sixel encodes at absolute pixel dimensions — the terminal paints
+	// one sixel pixel per device pixel rather than scaling into a cell
+	// box the way kitty does. Hand it the measured metrics so an image
+	// occupying N rows of layout is encoded N*cellHeight pixels tall and
+	// actually lands on those rows.
+	imgpkg.SetCellPixels(pxW, pxH)
 
 	// Wire the inline-image pipeline into the messages pane. SendMsg
 	// stays nil here because tea.NewProgram has not run yet; we re-call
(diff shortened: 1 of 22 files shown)


gammons:
Thanks @rfist !

PR #134, opened 2026-08-09
Fix reaction count fg lost after kitty image placement

### LINKED ISSUE #133: Own-reaction green styling lost on reaction counts when emoji_images is on (kitty/ghostty)
Noticed my own reactions weren't getting the green highlight on the count in ghostty, but the exact same message looked right in Terminal.app and in tmux. 

**Repro:** react to a message, view it in kitty or ghostty with `emoji_images = "on"` (the default). The count next to the emoji renders in the default fg instead of the own-reaction color. Same message in Terminal.app (or tmux without allow-passthrough) shows it green. Flipping `emoji_images = "off"` fixes it in kitty/ghostty too. slk 0.13.1 on macOS.

**What's happening:** the unicode-placeholder placement from `buildPlaceholderLines` (internal/image/kitty.go) encodes the image ID in the fg color and then ends with `\e[39m`. The pill in internal/ui/messages/model.go is styled as one string:

```go
pillText := fmt.Sprintf("%s%d", emojiStr, r.Count) // emojiStr = kitty placement
pills = append(pills, style.Render(pillText))       // ReactionPillOwn etc.
```

so the placement's trailing `\e[39m` resets the foreground right before the count gets printed and the style's fg never reaches it. The pill *background* survives (39 only touches fg).

**Possible fix:** re-assert the style's fg right after the placement in the pill loop (messages and thread panes), e.g. inject `ansi.Style{}.ForegroundColor(fg).String()` between the placement and the count when the image path was taken.

With emoji_images = "off"
Image
vs emoji_images = "on"

Image

achin90:
Fixes #133

The kitty unicode-placeholder encodes the image ID in the fg color and resets with `\e[39m` afterward, which wipes the pill style's foreground before the count digit prints. This re-asserts the style's fg between the placement and the count so the own-reaction green (and other pill colors) survive the image path.

Applied in both `internal/ui/messages/model.go` and `internal/ui/thread/model.go` (thread pane has its own copy of the pill loop).

Tested on ghostty (macOS, slk 0.13.1) — reaction counts render green again with `emoji_images = "on"`.

image

### DIFF
--- a/internal/ui/messages/model.go
+++ b/internal/ui/messages/model.go
@@ -2031,7 +2031,6 @@ func (m *Model) renderMessagePlain(msg MessageItem, width int, avatarStr string,
 					emojiStr = ":" + legacyName + ":"
 				}
 			}
-			pillText := fmt.Sprintf("%s%d", emojiStr, r.Count)
 			var style lipgloss.Style
 			if isSelected && m.reactionNavActive && i == m.reactionNavIndex {
 				style = styles.ReactionPillSelected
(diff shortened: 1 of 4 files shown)


gammons (commented):
Thanks @achin90 — the fix looks correct and the diagnosis (kitty placement's `\e[39m` wiping the pill fg) matches the code. The CI failure is an unrelated flake in `internal/slack/membership`, so a re-run should go green.

One request before merging: could you add a test covering this? The pill render loops in `internal/ui/messages` already have test infrastructure, and this is exactly the kind of regression that will silently come back. A test with a stubbed fetcher/placement asserting the count keeps the pill fg would do it.

Nit: the two copies diverge slightly (`strconv.Itoa` in messages vs `fmt.Sprintf` in thread) — keeping them identical helps since the loops are duplicated.

achin90:
Thanks for the review @gammons!

Had a lot of fun hacking on the project over the weekend... ended up with a stack of 14 changes to daily drive it at work today 😄

gammons:
thanks @achin90 !

Issue #50, opened 2026-05-29
Zellij support


jquievreux-matawan:
It seems that there are performance problems with Zellij. I can do tests, provide information if it helps support it. The support within Zellij would be a big plus for me 🙏🏼. 

 In any case, great project idea, thank you very much!

gammons:
Two things going on here, separated:

**1. The unresponsiveness — fixed in `174d896`.**

In zellij (and in tmux with `allow-passthrough = off`), slk's startup kitty graphics probe times out because the multiplexer doesn't forward APC `_G...` sequences to the host terminal. The old probe implementation spawned a goroutine that kept reading `os.Stdin` even after the timeout fired; that leaked goroutine then raced bubbletea's input loop for every byte the user typed, silently discarding ~95% of keystrokes (anything that isn't `0x1b`). That's why slk felt unresponsive but CPU was idle — input was being eaten before reaching the event loop.

The fix replaces the goroutine with a synchronous `poll(2)` + `read(2)` probe so nothing leaks. A regression test using `os.Pipe` asserts that a byte written after the probe timeout reaches the next reader intact. Will be in the next release.

**2. Why halfblock instead of real images in zellij.**

Same root cause as tmux-without-`allow-passthrough`: zellij parses our kitty graphics probe escape itself and drops it; kitty (the outer terminal) never sees it; no `;OK` reply; slk downgrades to halfblock. For tmux we wrap kitty escapes in the `\x1bPtmux;...\x1b\\` DCS passthrough envelope when `$TMUX` is set, and that round-trips correctly when the user opts into `allow-passthrough = on`. zellij has no equivalent envelope today.

If/when zellij adds an escape-passthrough mechanism (related context: zellij-org/zellij#775 — about keystroke passthrough, but in the same "things zellij intercepts rather than forwarding" family), adding zellij support to slk is a one-branch change next to the existing tmux wrapper at `internal/image/kitty.go:65`. Until then, your options for real kitty images are: use tmux + `allow-passthrough = on`, or run kitty without a multiplexer.

PR #138, opened 2026-08-12
Detect sixel support via DA1 instead of a TERM allowlist

### LINKED ISSUE #116: Someting is awry with sixel image support
I have this image in a slack channel:

Image

When I view the channel with slk, I see this:

Image

If I use `img2sixel` in the same terminal, I see:

Image

So it doesn't appear to be a terminal issue. For what it's worth, I'm running slk under tmux inside the `foot` terminal emulator. I see the same behavior if I run slk directly from the terminal without involving tmux.

srikary12:
Terminals that speak sixel were rendering the half-block mosaic because auto-detection only recognized sixel by name: TERM of "foot" or "mlterm", or TERM_PROGRAM of iTerm/WezTerm. Every other sixel-capable terminal -- xterm -ti vt340, DomTerm, toyterm, foot behind its foot-extra terminfo -- fell through to halfblock, which is what issue #116 reports as "blocky" output. Measuring the screenshots there confirms it: the blocks are flat and exactly cellW x cellH/2, the half-block signature, while img2sixel in the same terminals shows per-device-pixel detail.

Ask the terminal instead of guessing. ProbeSixel sends a Primary Device Attributes query (CSI c) and looks for attribute 4 in the reply, the same runtime check img2sixel and chafa use. Attributes are compared exactly so 14 / 24 / 41 don't false-positive.

The probe runs at startup only when detection already landed on halfblock, the config is auto (an explicit image_protocol=halfblock must stay half-block), and we're outside tmux/zellij -- tmux keeps sixel off by policy and zellij has no sixel path, so probing there would only risk paying the timeout for an answer we'd ignore. It also runs after a failed kitty probe, so that downgrade path gets a sixel chance too.

pollProbe is now parameterized by a scan function and shared with the kitty probe, keeping the poll(2)-based no-goroutine-leak path from issue #50. Detect also prefix-matches foot* / mlterm* so foot-extra works without needing the probe at all.

Fixes #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,9 +109,12 @@ tmux show -gv allow-passthrough
 Expected output: `on` (or `all`).
 
 If passthrough is off, `slk` detects this at startup and falls back to
-half-block rendering automatically — no config change needed. To force a
-specific renderer regardless of detection, set `image_protocol` in
-`config.toml` to `kitty`, `sixel`, `halfblock`, or `off`.
+half-block rendering automatically — no config change needed. Outside a
+multiplexer, terminals that don't do kitty graphics are asked whether
+they speak sixel (a DA1 query), so sixel-capable terminals get real
+pixels instead of the half-block mosaic. To force a specific renderer
+regardless of detection, set `image_protocol` in `config.toml` to
+`kitty`, `sixel`, `halfblock`, or `off`.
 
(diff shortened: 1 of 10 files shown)


gammons:
thanks @srikary12 !

PR #139, opened 2026-08-12
Fix skin-toned ZWJ emoji URL construction

achin90:
Skin-toned variants of ZWJ emoji — e.g. `:person_running_facing_right::skin-tone-2:` — built a CDN filename that 404s, so those reactions silently rendered with no image.

`ComposeSkinTonedCodepoints` appended the tone modifier to the end of the sequence:

```
1f3c3-200d-27a1-fe0f-1f3fc.png   404
```

A skin-tone modifier applies to the character immediately before it, so it belongs directly after the base codepoint, ahead of the ZWJ tail:

```
1f3c3-1f3fc-200d-27a1-fe0f.png   ✓
```

This is Unicode sequence semantics rather than a Slack-specific quirk; the CDN filenames just follow the codepoint order. For single-codepoint emoji like 👍, "after the base" and "at the end" are the same position, which is why ordinary skin tones were never affected.

## Tests

Two cases added — one at the codepoint level, one end-to-end as a CDN filename.

Worth noting for review: `TestComposeSkinTonedCodepoints` already existed, but every case used a single-codepoint base, where the two orderings are indistinguishable. The whole table passed identically before and after the fix. The new cases fail against the old implementation with the appended-modifier ordering.

### DIFF
--- a/internal/emoji/url.go
+++ b/internal/emoji/url.go
@@ -157,8 +157,17 @@ var skinToneCodepoints = [...]rune{
 
 // ComposeSkinTonedCodepoints decomposes a skin-toned shortcode name
 // (either Slack's "::skin-tone-N" or kyokomi's "_toneN" form), looks
-// up the base shortcode in the kyokomi codemap, and appends the
-// matching skin-tone modifier codepoint.
+// up the base shortcode in the kyokomi codemap, and inserts the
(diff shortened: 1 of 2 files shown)


gammons:
nice find, @achin90 !

PR #142, opened 2026-08-14
Use OSC 52 for clipboard writes

dannygreg:
## Summary
- route drag-selection and permalink copies through Bubble Tea's `tea.SetClipboard` OSC 52 command
- keep native clipboard access only for smart-paste reads, so copy still works when native initialization fails
- remove the obsolete native write failure path 
    - BubbleTea silently fails on clipboard write failures which is a bit of a shame 😕  

## Verification
- `go test ./... -race`
- `CGO_ENABLED=0 go build -trimpath ./cmd/slk`
- confirmed OSC 52 output contains the base64-encoded clipboard payload
- Tested successful copies through `zellij` and `herdr` in Ghostty on `macOS`

### DIFF
--- a/internal/ui/clipboard_write_test.go
+++ b/internal/ui/clipboard_write_test.go
@@ -1,0 +1,26 @@
+package ui
+
+import (
+	"reflect"
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+)
+
+func TestDefaultClipboardWriterUsesBubbleTeaOSC52(t *testing.T) {
+	const text = "copied through the terminal"
+	gotCmd := defaultClipboardWriter(text)
+	wantCmd := tea.SetClipboard(text)
+	if gotCmd == nil {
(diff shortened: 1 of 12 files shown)


internal/ui/clipboard_write_test.go
```diff
@@ -0,0 +1,26 @@
+package ui
+
+import (
+	"reflect"
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+)
+
+func TestDefaultClipboardWriterUsesBubbleTeaOSC52(t *testing.T) {
```
dannygreg, line 10: Not sure whether we want to keep this test… it doesn't add that much value.

gammons:
perfect.  thank you @dannygreg !

dannygreg:
I see a later run passed but should I be concerned about that CI failure? 😅 It doesn't look related to the clipboard change.

PR #144, opened 2026-08-17
fix: make thread parent message selectable

rfist:
## Problem

In the thread panel, the selection cursor could only land on **replies**. The parent (the first message that started the thread) was never selectable, so there was no way to react to it, copy its text/permalink, or run any message op against the message that opened the thread. Pressing `k` from the top reply had nowhere to go.

## Fix

Introduce a `parentSelected` (`-1`) sentinel for `Model.selected`:

- `SelectedReply()` returns the parent message when the sentinel is active, so every existing message op (react, copy, permalink, yank) targets it with no per-callsite changes.
- `MoveUp()` from the first reply now lands on the parent instead of stopping; `MoveDown()` moves back into the replies.
- Click and initial-selection paths recognise the parent row.
- The view highlights the parent row when it's selected.

## Tests

New `parent_selection_test.go` covers: parent becomes selectable, `SelectedReply()` resolves to the parent, up/down navigation crosses the parent boundary, and ops target the parent. `go test ./...` passes.

### DIFF
--- a/internal/ui/thread/model.go
+++ b/internal/ui/thread/model.go
@@ -89,6 +89,10 @@ type reactionHitRect struct {
 
 // Model represents the thread panel UI component.
 // It displays a parent message and its replies with cursor navigation.
+// parentSelected is the sentinel value of Model.selected meaning the
+// cursor sits on the parent message (the row above reply index 0).
+const parentSelected = -1
+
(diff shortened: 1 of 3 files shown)


gammons:
:ice_cream:

PR #145, opened 2026-08-17
fix: deliver live thread replies to panels showing a non-active channel

rfist:
## Problem

The open thread panel only received live replies when its thread belonged to the **active** channel. Reply routing lived inside the `if m.ChannelID == a.activeChannelID` branch of `reduceNewMessage`, so a thread panel opened from:

- the **Threads** view, or
- a **permalink** pointing into another channel

(neither of which touches `activeChannelID`) went stale. New replies didn't appear until the panel was closed and reopened, forcing a refetch.

## Fix

Hoist the reply-routing out of the active-channel branch and key it on the **panel's own identity** — its channel *and* its thread ts:

```go
if a.threadVisible &&
    m.ChannelID == a.threadPanel.ChannelID() &&
    m.Message.ThreadTS == a.threadPanel.ThreadTS() {
    a.threadPanel.AddReply(m.Message)
}
```

The new guard also adds the `ChannelID` match the old block lacked, so replies can't be mis-routed to a same-ts thread in a different channel, and there's no double-add (the old branch block is removed).

## Tests

`reducer_thread_live_test.go` covers a reply arriving for an open thread whose channel is not the active channel — it now lands in the panel. `go test ./...` passes.

### DIFF
--- a/internal/ui/reducer_send.go
+++ b/internal/ui/reducer_send.go
@@ -273,6 +273,19 @@ func reduceNewMessage(a *App, m NewMessageMsg) tea.Cmd {
 			mm.IncrementReplyCount(m.Message.ThreadTS, m.Message.TS)
 		}
 	}
+	// Route thread replies to the open thread panel keyed on the
+	// PANEL's identity (its channel + thread ts), not the active
+	// channel. The panel can be showing a thread from a non-active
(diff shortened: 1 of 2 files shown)


gammons:
nice find!

PR #147, opened 2026-08-18
fix: data race on the shared userNames map

piotrsynowiec:
## What breaks

`slk` can die outright with Go's `fatal error: concurrent map read and map write`. It is not recoverable — the runtime aborts the process, so the whole client disappears mid-session.

## Why

`resolveUserCached` memoised a resolved display name by writing it back into the shared `userNames` map. Its own doc comment stated the constraint:

```go
// UI-goroutine callers only — the map write is what
// lookupUserCached exists to avoid.
```

All four of its call sites violated it:

| Call site | Goroutine |
|---|---|
| `messageAuthor` via `fetchChannelMessages` | bubbletea command |
| `messageAuthor` via `fetchOlderMessages` | bubbletea command |
| `messageAuthor` via `fetchThreadReplies` | bubbletea command |
| `rtmEventHandler.OnMessage` | WebSocket reader |

The UI goroutine reads that same map while rendering, so every resolved name raced it.

The window is narrow and data-dependent: the write only fires on a **cache hit for a name not yet memoised in this session**. That makes it common on a warm cache and absent on a cold one — the opposite of what a fresh test run exercises, which is likely why it went unnoticed.

## The fix

Point the call sites at `lookupUserCached`, which does not write, and delete `resolveUserCached` rather than leave it unused. A function whose stated contract no caller satisfies is a trap for the next person.

Nothing is lost by dropping the memoisation: it bought one indexed SQLite point lookup, against a map the resolver repopulates on the UI goroutine anyway.

## Testing

`cmd/slk/message_author_race_test.go` drives `messageAuthor` from a background goroutine while the main goroutine reads the map, and is meant to be run under `-race`.

The test was validated against the defect, not just against the fix: reintroducing the write makes `-race` report it. A race test that has never failed proves nothing, so this seemed worth confirming.

Full suite passes, including `go test -race ./cmd/slk/`.

---

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -2745,19 +2745,16 @@ func lookupUserCached(userID string, userNames map[string]string, db *cache.DB) 
 	return "", false
 }
 
-// resolveUserCached is lookupUserCached plus map memoization: a DB hit
-// is written back to userNames so subsequent lookups skip SQLite.
-// UI-goroutine callers only — the map write is what lookupUserCached
-// exists to avoid. Returns ("", false) when the user is unknown —
-// caller is expected to fall back to userID-as-name and enqueue an
-// async lookup via wctx.UserResolver.Request.
-func resolveUserCached(userID string, userNames map[string]string, db *cache.DB) (string, bool) {
-	name, ok := lookupUserCached(userID, userNames, db)
-	if ok {
-		userNames[userID] = name
-	}
-	return name, ok
-}
+// resolveUserCached used to sit here: lookupUserCached plus a
+// write-back into the shared userNames map, documented "UI-goroutine
+// callers only". Every one of its call sites was off the UI goroutine —
+// three fetch paths running as bubbletea commands and rtmEventHandler
+// .OnMessage running on the WebSocket goroutine — so the map write
+// raced the UI's reads on every resolved name. A function whose stated
+// contract nothing satisfies is a trap, not an optimisation, so it is
+// gone rather than merely unused; the memoisation it bought was one
+// SQLite point lookup, and the resolver patches names into the models
+// anyway.
 
 // resolveUser ensures we have the display name and avatar for a user.
 // If the user is unknown, fetches their profile from Slack on demand.
@@ -2895,7 +2892,7 @@ func resolveDMNames(wctx *WorkspaceContext, db *cache.DB, avatarCache *avatar.Ca
 // and the MessageItem are keyed on so the avatar pipeline can attach.
 func messageAuthor(m slack.Message, userNames map[string]string, db *cache.DB, router *workspaceRouter) (userID, userName string) {
 	if m.User != "" {
-		name, ok := resolveUserCached(m.User, userNames, db)
+		name, ok := lookupUserCached(m.User, userNames, db)
 		if !ok {
 			name = m.User
 			if router != nil {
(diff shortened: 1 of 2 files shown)


gammons:
The bug is real and it's the most serious thing in the current PR queue — thanks for finding it. It's also **worse than your description claims**, in your favour.

Every call site of `resolveUserCached` is off the UI goroutine, despite the doc comment at `main.go:2771` saying "UI-goroutine callers only":

| call site | goroutine |
|---|---|
| `messageAuthor` (`main.go:2919`, `2933`) via `fetchChannelMessages` | bubbletea cmd (`reducer_channels.go:391`) |
| via `fetchOlderMessages` | bubbletea cmd (`app.go:1490-1492`) |
| via `fetchMessagesAround` | bubbletea cmd (`reducer_links.go:103-105`) |
| via `fetchThreadReplies` | bubbletea cmd (`reducer_threads.go:82`) |
| `rtmEventHandler.OnMessage` (`main.go:4058`) | WebSocket reader |

And the description says "the write only fires on a cache hit for a name not yet memoised in this session" — that's not right. `resolveUserCached` writes back on *every* `ok`, including a hit that came straight out of `userNames` itself (`main.go:2776-2779`). So it fires on essentially every message with a known author. Not a narrow window at all.

I reproduced it by restoring the write-back under your test:

```
fatal error: concurrent map read and map write
WARNING: DATA RACE
  Previous write at 0x00c00027a138 by goroutine 48:
    resolveUserCachedOLD() main.go:2896
    messageAuthor()        main.go:2903
```

A `fatal error` that kills the whole client mid-session is the highest-value class of bug in a TUI, and removing the writer rather than adding a lock is the right call on a hot path.

**Two things before I merge.**

**1. The fix is incomplete — there's a second unsynchronised writer.** `resolveUser` writes `userNames[userID] = name` at `main.go:2827`, and its caller `resolveDMNames` is launched as a bare `go resolveDMNames(...)` at `main.go:2152`. Same map, same UI-goroutine readers, plus the WS-goroutine reads at `main.go:3988` and `main.go:4002`. You fixed 4 of 5 off-UI writers; please fix or explicitly scope out the fifth.

**2. `TestMessageAuthor_DoesNotMutateSharedNameMap` passes on the unfixed code.** I verified — restoring `resolveUserCachedOLD` gives `--- PASS`. With `db == nil` the write-back only ever re-writes an existing key, so `len(userNames)` never changes and the unknown-user branch is never reached. Its comment claims "the plain run still catches a regression via the explicit map-mutation assertion below," which isn't true. Either delete it or give it a real `*cache.DB` with a user row absent from the map — that's the only configuration where the assertion has teeth.

`TestMessageAuthor_SafeUnderConcurrentReads` is the one that earns this PR. I confirmed it fails on pre-fix code with both a `-race` report and a hard runtime `fatal error`. Keep that one.

Smaller:

- `main.go:215` and `main.go:431` still reference `resolveUserCached` by name — dangling doc references to a deleted function.
- The 10-line tombstone comment at `main.go:2748-2757` explaining what used to be there is diff noise; git history covers it. I'd drop it.
- `TestMessageAuthor_FallsBackToTheID`'s comment says "a message with neither a user nor a bot id" but it passes `User: "U-UNRESOLVED"`. Also overlaps `bot_identity_test.go:29`.
- Please correct the description — the write fires on every cache hit, not only un-memoised ones.

Your CI lint failure is **not your fault**: it's an old golangci-lint panicking under go1.27 (`could not import math/rand/v2 ... (typecheck)`), fixed on main by 6d39fe5. Rebase and it clears. Your `test` job passed.

gammons:

PR #148, opened 2026-08-18
fix: --add-workspace can write a config slk refuses to load

piotrsynowiec:
## What breaks

Running `slk --add-workspace` a second time can leave `config.toml` in a state `slk` refuses to load, so the client stops starting until the file is hand-edited. A picker default turns into a broken install.

## Why

Two things combine:

1. The picker **pre-selected every workspace**, including ones already configured. Re-running `--add-workspace` to add one workspace therefore re-added all of them by default.
2. The config writer de-duplicated on the **slug**, not the `team_id`. A second pass wrote a second `[workspaces.<slug>-2]` block carrying the same `team_id`.

`config.Load` rejects duplicate team IDs, so the very next launch fails. Worse, the failure is self-perpetuating: `--add-workspace` was one of the few things still runnable, and running it again appended a *third* block to a file `slk` already would not start on.

## The fix

Two changes, deliberately independent so neither depends on the other holding:

**The writer skips a `team_id` it already has.** Keying the guard at the writer rather than on callers remembering to filter makes it hold no matter who calls it, including callers not yet written.

**The picker pre-selects only unconfigured workspaces**, and labels the rest. Re-selecting one is still allowed — that is how you refresh a token — it just no longer happens by accident.

## One implementation note worth flagging

`configuredTeamIDs` decodes `config.toml` directly instead of going through `config.Load`.

This is deliberate. `Load` validates, and validation is exactly what fails on a config that already carries a duplicate. Routing the guard through `Load` would make it useless in the one case it exists for — recovering a file that is already broken — and a third run would keep appending. There is a test that constructs that broken state on purpose to pin this down.

## Testing

`cmd/slk/add_workspace_config_dedupe_test.go` covers the duplicate-`team_id` guard, the slug-collision path, and recovery from an already-broken config.

Full suite passes.

---

### DIFF
--- a/cmd/slk/add_workspace_config.go
+++ b/cmd/slk/add_workspace_config.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	toml "github.com/pelletier/go-toml/v2"
+
 	"github.com/gammons/slk/internal/config"
 )
 
@@ -31,7 +33,20 @@ func uniqueSlug(base string, existing map[string]bool) string {
 // team_id set, prefixed by a "# <teamName>" comment line. The file
 // is created if it does not exist. Existing content is preserved
 // verbatim (textual append, not TOML re-marshal).
+//
+// Writing a second block for a team_id that already has one is a no-op.
+// The slug is what uniqueSlug de-duplicates, and two different slugs
+// can name the same workspace — re-running --add-workspace with an
+// already-configured workspace selected produced exactly that, and
+// config.Load then rejects the file, so slk refuses to start until the
+// user hand-edits it. Keying the skip on team_id rather than on the
+// caller remembering to filter makes the writer safe no matter who
+// calls it.
 func appendWorkspaceConfigBlock(configPath, slug, teamID, teamName string) error {
+	if teamID != "" && configuredTeamIDs(configPath)[teamID] {
+		return nil
+	}
+
 	var existing []byte
 	if data, err := os.ReadFile(configPath); err == nil {
 		existing = data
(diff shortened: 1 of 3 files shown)


gammons:
Nice fix — keying the guard on `team_id` at the writer and decoding the TOML directly (so recovery from an already-broken config works) is the right call, and the tests cover the important states.

One thing before merge, @piotrsynowiec: the new picker strings are in Polish (`— już dodany`, `Wszystkie workspace'y…`, `Nowe są zaznaczone; spacja przełącza…`) while the rest of the UI — including this same dialog's title and the surrounding onboarding flow — is English. Could you switch these to English for consistency?

gammons:
Real bug, correctly diagnosed, and the approach is sound — but there's a blocker.

**The bug is confirmed, and it's self-perpetuating.** `onboarding.go:49-52` pre-selects every workspace, `onboarding.go:106` de-dupes on slug only so a second run yields `acme-2`, `appendWorkspaceConfigBlock` appends a block with the same `team_id`, and `internal/config/workspaces.go:74-77` then refuses to load it. And because `existingSlugs` routes through `config.Load` (`add_workspace_config.go:68-71`), which now errors and returns an empty set, `uniqueSlug` hands back the *base* slug — so a third run writes a duplicate TOML table key, a parse error stacked on top of a validation error. Bricking someone's install from the happy path of the onboarding command is about as bad as first-run UX gets.

Two independent guards (idempotent writer + smarter picker default) is good defensive design, and your justification for bypassing `config.Load` in `configuredTeamIDs` is correct — `Load` fails on precisely the file the guard exists to recover.

**Blocker: the PR contains untranslated Polish UI strings.**

```go
label += "  — już dodany"
fmt.Println(dimStyle.Render("  Wszystkie workspace'y z aplikacji desktopowej są już dodane."))
fmt.Println(dimStyle.Render("  Zaznacz któryś, żeby odświeżyć jego token, albo wyjdź (Ctrl+C)."))
Description("Nowe są zaznaczone; spacja przełącza, enter zatwierdza.")
```

The last one *replaces* the existing English string at `onboarding.go:66`. slk is English-only with no i18n layer — `grep -P '[ąćęłńóśźżĄĆĘŁŃÓŚŹŻ]'` across all non-test Go source on main returns zero matches. These need to go back to English, including restoring `onboarding.go:66`.

**Second issue: the guard misses the legacy config format it needs to cover.** `configuredTeamIDs` only reads the `team_id` *field* and ignores team-ID-*keyed* blocks, which `config.Load` explicitly supports (`internal/config/workspaces.go:66-68`, `isTeamIDKey` at `workspaces.go:19`). Reproduced:

```
configuredTeamIDs = map[]      <- legacy [workspaces.T011B427MLP] invisible
resulting config:
  [workspaces.T011B427MLP]
  sidebar_width = 30
  [workspaces.acme]
  team_id = "T011B427MLP"
GUARD MISSED LEGACY KEY -> config.Load now fails:
  workspaces "T011B427MLP" and "acme" both reference team_id "T011B427MLP"
```

So a user on a legacy config hits the exact bug this PR fixes. The picker's `already` map has the same gap. Two lines: also treat a key matching `^[TE][A-Z0-9]{6,}$` as a configured team ID.

Smaller:

- `filepath.Join(xdgConfig(), "config.toml")` is now built twice in `onboarding.go`; `newCount` duplicates `len(chosen)`.
- `TestConfiguredTeamIDs_ReadsSingleQuotedValues` tests `go-toml`, not your code. I'd drop it.
- `TestConfiguredTeamIDs_WorksOnAConfigThatFailsValidation` is well chosen — it pins the design decision your description flags. Keep that one.
- Nothing covers the `onboarding.go` picker change. `addWorkspace` isn't testable as written so I won't insist, but the description shouldn't imply coverage it doesn't have.
- `existingSlugs`'s own failure mode (empty set on a broken config, leading to a duplicate table key) is still there — now masked by the new guard rather than fixed.

Your CI lint failure is **not your fault**: old golangci-lint panicking under go1.27, fixed on main by 6d39fe5. Rebase and it clears.

gammons:

PR #149, opened 2026-08-18
fix: opening a thread too narrow to show gave no feedback

piotrsynowiec:
## What breaks

Open a thread in a terminal too narrow for the panel and nothing appears to happen. The status bar still reads `> Thread`, so you are told you are in a thread while looking at no thread — and there is no way to tell a failed open from a panel that simply did not fit.

## Why

The thread panel needs a 35% width share. Below that, the layout sets `threadVisible = false` and returns, but the status-bar flag that renders `> Thread` was left set. Two pieces of state that must agree could disagree.

## The fix

Clear the status-bar flag alongside the visibility, so the two cannot drift apart, and surface a toast explaining that the terminal is too narrow — the actionable part, since widening the window fixes it.

The toast fires only on the transition into the hidden state, so a session spent at a narrow width does not repeat it on every frame.

## Testing

`internal/ui/thread_autohide_test.go` covers the narrow-open path, the status-bar flag staying in sync, and the toast firing once rather than per frame.

Full suite passes.

---

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -2600,7 +2600,16 @@ func (a *App) View() tea.View {
 	// thread pane at its minimum.
 	frame := a.layout.Compute(a.width, a.height, a.workspaceRail.Width(), a.sidebar.Width(), a.sidebarVisible, a.threadVisible)
 	if frame.ThreadAutoHidden {
+		// Say so. Dropping threadVisible silently left the status bar
+		// still reading "> Thread" while no thread panel was on
+		// screen: the user opened a thread, was told they were in one,
+		// and had nothing to look at — with no way to tell whether the
+		// open had failed or the panel was merely too wide to fit.
+		if a.threadVisible {
+			a.statusbar.SetToast("Panel wątku nie mieści się — ctrl+b chowa sidebar")
+		}
 		a.threadVisible = false
+		a.statusbar.SetInThread(false)
 		if a.focusedPanel == PanelThread {
 			a.focusedPanel = PanelMessages
 		}
--- a/internal/ui/thread_autohide_test.go
+++ b/internal/ui/thread_autohide_test.go
@@ -1,0 +1,71 @@
+package ui
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/gammons/slk/internal/ui/messages"
+	"github.com/gammons/slk/internal/ui/sidebar"
+)
+
+func appWithOpenThread(t *testing.T, width int) *App {
+	t.Helper()
+	a := NewApp()
+	a.width, a.height = width, 40
+	a.activeTeamID, a.activeChannelID = "T1", "C1"
+	a.sidebar.SetItems([]sidebar.ChannelItem{{ID: "C1", Name: "newsletter", Type: "channel"}})
+	a.messagepane.SetMessages([]messages.MessageItem{
+		{TS: "1750000000.000100", UserName: "Michał", Text: "Cześć", ReplyCount: 1},
+	})
+	a.focusedPanel = PanelMessages
+	_ = a.View()
+	a.handleEnter() // opens the thread
+	return a
+}
+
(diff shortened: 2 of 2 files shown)


gammons:
Good catch — clearing `SetInThread(false)` alongside the visibility closes the state drift, and firing the toast only on the transition (since `threadVisible` is cleared) is the right shape.

One thing before merge, @piotrsynowiec: the toast is in Polish (`Panel wątku nie mieści się — ctrl+b chowa sidebar`) but the rest of the UI's toasts and status messages are English (`Invalid snooze duration`, `Copied permalink`, etc.). Could you switch it to English? The test asserts on the Polish text too, so it'll need updating alongside.

gammons:
There are two changes in here with opposite verdicts.

**The one-line status-bar fix is correct and I want it.** `openThreadPanel` sets both flags (`app.go:1587-1588`) but `View()` clears only `threadVisible` (`app.go:2694-2699`), so `statusbar/model.go:226` keeps rendering `" ... > Thread "` after an auto-hide. Every other path pairs them correctly (`app.go:1734`, `app.go:1770`); auto-hide is the one that doesn't. `a.statusbar.SetInThread(false)` is exactly right.

**The toast should not land.** Three reasons, all verified on your branch:

**1. The advice it gives doesn't work.** The message says `ctrl+b` hides the sidebar so the panel will fit. It won't:

```
threadVisible after widening:                false
TOAST STILL ON SCREEN after widening
threadVisible after ctrl+b (sidebar hidden): false
```

`a.threadVisible = false` is destructive and nothing ever sets it back, so `Compute` runs with `threadVisible=false`, `autoHidden` is never re-evaluated, and the panel does not return. We'd be telling users to press a key that provably does nothing. The real root cause is that auto-hide destroys intent instead of deriving presentation from it — a separate `threadAutoHidden` field with `threadVisible` left intact would make the panel reappear on resize *and* make the message true.

**2. The toast never clears.** `SetToast`'s own contract (`statusbar/model.go:186-188`) says callers must clear it, typically via a `tea.Tick` delivering `CopiedClearMsg`. But `View()` returns `tea.View`, not `tea.Cmd` — it structurally cannot schedule that tick. Confirmed above: still on screen after widening, and it stays until some unrelated toast overwrites it. Every other toast in the codebase is emitted from `Update` as a `ToastMsg` (`reducer_search.go:101`, `reducer_search.go:165`, `mode_workspace_search.go:57`) precisely so it can be cleared.

**3. It has side effects inside `View()`.** Mutating `a.threadVisible` there is a pre-existing wart, but `SetToast` calls `m.dirty()` (`statusbar/model.go:191`), so this invalidates the render cache from inside the render.

**Also blocking: the UI string is in Polish.** `"Panel wątku nie mieści się — ctrl+b chowa sidebar"`, plus the test asserting on `"nie mieści"`. slk is English-only with no i18n layer. The test fixtures are Polish too (`"Michał"`, `"Cześć"`) where the rest of the package uses `alice`/`bob`.

On tests: `TestThreadAutoHidden_StatusBarStopsClaimingThread` is legitimate and would fail without the fix. `TestThreadAutoHidden_WideTerminalKeepsThePanel` asserts `strings.Contains(a.View().Content, "Thread")` — bare `"Thread"` matches almost anything on screen, so it's near-worthless. And the description claims tests cover "the toast firing once rather than per frame"; no such test exists, and nothing covers the post-resize behaviour, which is where this actually falls over.

**What I'd like:** cut this down to `a.statusbar.SetInThread(false)` at `app.go:2701` and I'll merge it. If you want narrow-terminal feedback, it's a separate PR that needs the non-destructive auto-hide fix first, emits a `ToastMsg` from `Update` so it can be cleared, and says something true.

Your CI lint failure is **not your fault** — old golangci-lint panicking under go1.27, fixed on main by 6d39fe5.

gammons:

PR #150, opened 2026-08-18
fix: unread dot that can never be cleared when the message is unfetchable

piotrsynowiec:
## What breaks

A channel shows an unread dot in the sidebar while the message pane reads **"No messages yet"**, and nothing the user does clears it. Opening the channel does not help. The dot is permanent.

## Why

`client.counts` decides `has_unreads` by comparing a channel's `latest` against `last_read` — and it does so even when `latest` is no longer retrievable. On a retention-limited plan, a dormant channel gives these three answers at the same time:

```
client.counts          has_unreads: true
                       last_read:   1718264828.618929   (Jun 2024)
                       latest:      1776079647.792389   (Apr 2026)

conversations.history  ok: true, messages: [], is_limited: true

conversations.info     properties.is_dormant: true
```

Slack is not wrong, and neither is `slk` for storing the flag — the unread message genuinely exists, it has simply aged out of the fetchable window. It can never be displayed, and therefore never read.

The problem is that **every** mark-read path was gated on having at least one message:

| Location | Gate |
|---|---|
| `cmd/slk/main.go` — `Fetch` | `if len(msgItems) > 0` |
| `internal/ui/reducer_channels.go` — Tier 1 | `if len(cached) == 0 { return nil, false }` |
| `markChannelReadAsync` | early return on `ts == ""` |

With no message there is no timestamp to mark at, so `conversations.mark` was never sent.

## The fix

`channelMarkTS` picks the mark timestamp from the nil-vs-`[]` contract `fetchChannelMessages` already documents and honours:

| Fetch result | Meaning | Action |
|---|---|---|
| `nil` | history call **failed** | mark nothing — we know nothing |
| `[]` + `has_unread` | Slack has nothing to show us | mark at Slack's `latest` |
| `[]` + already read | ordinary empty channel | do nothing |
| non-empty | normal case | mark at newest message (unchanged) |

### Why `latest` and not `time.Now()`

This is the part worth reading, because the obvious implementation is wrong in a way that is easy to miss.

`last_read_ts` does not only drive the unread dot. It is also what the sidebar's staleness filter reads: `IsStale` hides a channel once `now - last_read_ts` exceeds `hide_inactive_after_days` (30 by default), and it treats an unread channel as *never* stale.

So a dormant channel is pinned to the sidebar by `has_unread`. Marking it at the wall clock clears the dot — and then pins it right back for the full threshold window, because it now claims the user just read a channel whose newest message is months old. The symptom moves; the channel still will not go away.

Marking at Slack's real `latest` clears the dot **and** leaves the channel its true age, so the staleness filter ages it out on its own. On the workspace this was found on, that is 127 days past the threshold.

I found this the hard way: the first version of this patch used `time.Now()`, and the channel stayed in the sidebar with no dot. `TestChannelMarkTSStaysStale` is the regression guard.

### Where `latest` comes from

`client.counts` is the only endpoint that reports it. `conversations.info` omits the field entirely, and `conversations.history` cannot return a message that has aged out of the retention window — which is the whole situation. `UnreadInfo` now carries it.

The lookup is a network call, so it fires **only** on the branch that needs it: an authoritative empty fetch on a channel flagged unread. Two tests pin that it does not fire on the ordinary paths.

When `latest` cannot be determined, the channel is left unmarked. A dot that clears on the next open is a smaller error than a read state written from a guess.

### Why it marks through Slack

Boot's `ReplaceWorkspaceReadState` re-applies `client.counts` as authoritative, so a local-only clear would light the dot again on the next start. It has to be `conversations.mark`.

The `nil` case is what keeps this safe: a transient network failure returns `nil`, never `[]`, so a failed fetch can never mark a channel with real unread messages as read.

## Testing

`cmd/slk/channel_mark_ts_test.go` covers every branch, both "no lookup on the ordinary paths" guards, and the staleness regression above.

Verified end to end against a live workspace, including resetting the channel's read state back to the broken condition and reproducing the whole cycle: the dot clears, `client.counts` then reports `has_unreads: false`, and the channel drops out of the sidebar.

Full suite passes.

---

### DIFF
--- a/cmd/slk/channel_mark_ts_test.go
+++ b/cmd/slk/channel_mark_ts_test.go
@@ -1,0 +1,146 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/gammons/slk/internal/cache"
+	"github.com/gammons/slk/internal/ui/messages"
+)
+
+// TestChannelMarkTS pins the fetch outcomes that decide whether a
+// channel is marked read, and at which timestamp.
+func TestChannelMarkTS(t *testing.T) {
+	// Slack's `latest` for a dormant channel: real, long past, and
+	// naming a message conversations.history will not return.
+	const slackLatest = "1776079647.792389"
+	latest := func() string { return slackLatest }
+
+	tests := []struct {
+		name   string
+		items  []messages.MessageItem
+		state  cache.ReadState
+		lookup func() string
+		want   string
+	}{
+		{
+			name:   "fetch failed: never mark, even when flagged unread",
+			items:  nil,
+			state:  cache.ReadState{HasUnread: true},
+			lookup: latest,
+			want:   "",
+		},
+		{
+			name:   "empty and already read: nothing to do",
+			items:  []messages.MessageItem{},
+			state:  cache.ReadState{HasUnread: false},
+			lookup: latest,
+			want:   "",
+		},
+		{
+			// The dormant / retention-limited channel. Marking at
+			// Slack's own `latest` clears the unread without
+			// fabricating recency the channel never had.
+			name:   "empty but flagged unread: mark at Slack's latest",
+			items:  []messages.MessageItem{},
+			state:  cache.ReadState{LastReadTS: "1718264828.618929", HasUnread: true},
+			lookup: latest,
+			want:   slackLatest,
+		},
+		{
+			name:   "empty, unread, but latest unknown: do not guess",
+			items:  []messages.MessageItem{},
+			state:  cache.ReadState{HasUnread: true},
+			lookup: func() string { return "" },
+			want:   "",
+		},
+		{
+			name:   "empty, unread, no lookup wired: do not mark",
+			items:  []messages.MessageItem{},
+			state:  cache.ReadState{HasUnread: true},
+			lookup: nil,
+			want:   "",
+		},
+		{
+			name:   "has messages: mark at the newest one",
+			items:  []messages.MessageItem{{TS: "100.000001"}, {TS: "200.000002"}},
+			state:  cache.ReadState{HasUnread: true},
+			lookup: latest,
+			want:   "200.000002",
+		},
+		{
+			name:   "has messages while already read: still marks newest",
+			items:  []messages.MessageItem{{TS: "300.000003"}},
+			state:  cache.ReadState{HasUnread: false},
+			lookup: latest,
+			want:   "300.000003",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := channelMarkTS(tt.items, tt.state, tt.lookup); got != tt.want {
+				t.Errorf("channelMarkTS() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+// TestChannelMarkTSNoLookupWhenMessagesPresent guards the cost side:
+// the `latest` lookup is a network call, and it must not fire on the
+// ordinary path where the newest fetched message already answers the
+// question.
+func TestChannelMarkTSNoLookupWhenMessagesPresent(t *testing.T) {
+	called := false
+	items := []messages.MessageItem{{TS: "400.000004"}}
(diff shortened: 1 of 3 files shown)


gammons:
Real bug, correct diagnosis, right timestamp choice — but the plumbing needs work.

**Confirmed the bug.** Every mark-read path is gated on having messages: `cmd/slk/main.go:1465`, `reducer_channels.go:405`, and `markChannelReadAsync` early-returning on `ts == ""` (`main.go:3094`). And there's no escape hatch — `markUnreadOfSelected` needs a selected message and `command.go` has no mark-read command. Since `sidebar/model.go:337` means `IsStale` never hides an unread channel, the channel also stays pinned in the sidebar and keeps catching `a`/`A`. Permanent dot, no workaround. On free-tier or 90-day-retention workspaces that's common, and it's exactly the papercut a keyboard-first client can't ship.

The `latest` field is real too — the repo's own captured contract (`internal/slack/testdata/phase2-api-contracts.json`) shows `client.counts` returning `latest` on `channels`, `mpims` and `ims`. The `internal/slack/client.go` half of this diff is correct and decode-safe. And marking at the real `latest` rather than `time.Now()` is the right call (`sidebar/staleness.go:42`).

**Main issue: it re-fetches data we already have, on a loop.**

`client.counts` is already called at `cmd/slk/bootstrap_adapters.go:100` and `cmd/slk/reconnect_sync.go:129`. Both now receive `UnreadInfo.Latest` and throw it away. Then `latestFromCounts` fires a *third* full-workspace `client.counts` round-trip on channel open. And if `latest` comes back empty or `conversations.mark` is rejected, `HasUnread` stays true — so that round-trip repeats on every subsequent open of that channel, forever.

Persisting `latest` alongside `last_read_ts`/`has_unread` (`internal/cache/channels_read_state.go:10`) at the two call sites we already have is both cheaper and simpler.

**Second: `latestFromCounts` ignores the package's own testability pattern.** It takes a concrete `*slackclient.Client`, while `cmd/slk/reconnect_sync.go:37` defines a one-method interface for exactly this call, in the same package. Consequence: it's untestable, and untested.

**Third: the tests cover the one thing that couldn't have been wrong.** 146 test lines for a 12-line pure function, and nothing else. No test for `latestFromCounts`, and no test that the `Fetch` closure at `main.go:1454` marks anything — the actual behavior change is uncovered. `TestChannelMarkTSNoLookupWhenMessagesPresent` and `...WhenAlreadyRead` duplicate table rows 6 and 2 with a call-counter bolted on. `TestChannelMarkTSStaysStale` duplicates row 3 and its `if got == now` branch is dead — with the lookup injected, `got` can only be `slackLatest` or `""`.

(Minor: `got <= "1718264828.618929"` is a lexicographic compare on a Slack ts. Works here by digit-count coincidence, not by design.)

**Also:** the comment at `main.go:1454-1456` still says the channel is marked "up to now." It isn't — that's left over from the revision your description says was wrong.

**What I'd like:**

1. Persist `UnreadInfo.Latest` at `bootstrap_adapters.go:100` and `reconnect_sync.go:129`, read it from the db in the `Fetch` closure. Drop `latestFromCounts`, or keep it as a fallback behind a narrow interface like `reconnect_sync.go:37`.
2. Fix the stale `main.go:1454-1456` comment.
3. Add a test for the `Fetch` wiring: empty + unread marks at `latest`; empty + unread with a nil fetch doesn't mark. Drop the dead `now` assertion and fold the two duplicate call-counter tests into the table.
4. Optional: `channelMarkTS` is pure read-state *policy* with no I/O, and that policy already has a home in `internal/` (`cache/channels_read_state.go`, `reducer_channels.go`). Moving it there would also let Tier 1 at `reducer_channels.go:405` share it later. `cmd/slk` isn't off-limits by convention, so take or leave.

Also: the 32-line doc comment on a 12-line function restates the PR description verbatim. House style tolerates long comments but this is at the far end.

Your CI lint failure is **not your fault** — old golangci-lint panicking under go1.27, fixed on main by 6d39fe5. Your `test` job passed.

gammons:

PR #151, opened 2026-08-18
fix: clicking a sidebar section header did nothing

piotrsynowiec:
## What breaks

Clicking a section header in the sidebar does nothing. The row highlights, so it looks like it registered — and then nothing happens. Collapsing or expanding a section is keyboard-only.

It shows up fastest on the **Channels** section, which ships collapsed by default: clicking it open is the obvious first move, and it silently fails.

## Why

The sidebar has two activation paths, and they had drifted apart.

`handleEnter` dispatches on three row kinds:

```go
if a.sidebar.IsThreadsSelected() { ... }      // Threads row
if a.sidebar.ToggleCollapseSelected() { ... } // section header
item, ok := a.sidebar.SelectedItem()          // channel
```

`reduceMouseClick` handled only two of them — channels and Threads:

```go
if item, ok := a.sidebar.ClickAt(sidebarY); ok { ... }  // channel
if a.sidebar.IsThreadsSelected() { ... }                // Threads row
return nil                                              // everything else
```

`ClickAt` returns `ok=false` for *every* non-channel row, and the mouse reducer treated that as "nothing here" after a single special case for Threads. It had already moved the selection to the clicked row — which is what draws the highlight and makes the row look live — and then returned nil.

## The fix

Add the missing branch, in `handleEnter`'s order so the two paths read as the same decision rather than two lists that happen to overlap.

## Testing

The test sweeps **every rendered row** and asserts click and Enter agree on each, instead of asserting one row kind at a time.

That is the deliberate part. A per-row-kind test would have had exactly the blind spot that caused this: the row that broke is the row nobody thought to write a case for. The sweep fails on the section-header row today, and will fail again for the next row kind wired into one path and not the other.

Verified by reverting the fix and confirming the test fails on the defect, so it fails on the bug rather than merely passing on the patch.

Full suite passes.

---

### DIFF
--- a/internal/ui/reducer_mouse.go
+++ b/internal/ui/reducer_mouse.go
@@ -5,18 +5,18 @@
 // Owns the two remaining mouse Update arms that act as multi-panel
 // routers:
 //
-//   tea.MouseWheelMsg  - viewport scroll for the panel under the
-//                        cursor (sidebar, messages pane / threads
-//                        view, thread panel). Decoupled from j/k
-//                        selection. Triggers maybeFetchOlderHistory
-//                        on the messages pane when the viewport
-//                        hits the top.
-//   tea.MouseClickMsg  - panel-router: workspace rail (switch
-//                        workspace), sidebar (channel select /
-//                        threads view), messages pane (threads
-//                        click / reaction hit-test / image hit-test
-//                        / drag begin), thread panel (reaction
-//                        hit-test / drag begin).
+//	tea.MouseWheelMsg  - viewport scroll for the panel under the
+//	                     cursor (sidebar, messages pane / threads
+//	                     view, thread panel). Decoupled from j/k
+//	                     selection. Triggers maybeFetchOlderHistory
+//	                     on the messages pane when the viewport
+//	                     hits the top.
+//	tea.MouseClickMsg  - panel-router: workspace rail (switch
+//	                     workspace), sidebar (channel select /
+//	                     threads view), messages pane (threads
+//	                     click / reaction hit-test / image hit-test
+//	                     / drag begin), thread panel (reaction
+//	                     hit-test / drag begin).
 //
 // Free reducer (not controller-absorbed) because both arms route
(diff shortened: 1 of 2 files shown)


gammons:
Reviewed. Merging this.

The gap is real and I confirmed it against both dispatch paths: `handleEnter` (`internal/ui/app.go:1517`) handles three sidebar row kinds — `IsThreadsSelected`, `ToggleCollapseSelected`, `SelectedItem` — while `reduceMouseClick` (`internal/ui/reducer_mouse.go:223`) only handled two. And since `ClickAt` moves the cursor before returning `false` (`internal/ui/sidebar/model.go:1673-1676`), the row visibly highlights and then does nothing, which is worse than ignoring the click outright. Channels ships collapsed, so this is one of the first clicks a new user makes.

The test earns its place — I removed just the added branch and it fails with `row y=2: click and Enter left different sidebar state`. Framing it as a sweep comparing click vs Enter is the right call for a bug that's two copies of one decision drifting apart.

Also worth noting the 15-line comment reindent isn't churn: `gofmt -d internal/ui/reducer_mouse.go` produces exactly that hunk on main. You fixed a formatting drift while you were in there. (I'm enforcing gofmt in CI in #171 so that stops leaking into feature diffs.)

Two non-blocking follow-ups I'll file rather than hold this up for:

- `maxRows = 12` in the test exceeds the fixture's rendered row count, so rows past the content leave the cursor at its default and both paths trivially emit `ThreadsViewActivatedMsg`. That makes the `activated == 0` guard weaker than it reads. Tightening `maxRows` to the rendered count would sharpen it.
- The real underlying issue is that `ClickAt` reports only a bool, so callers have to consult *stale* selection state to find out what was hit. Clicking a blank filler row toggles whatever was previously selected. Main already has that behavior via `IsThreadsSelected`, so it's not a regression here, but having `ClickAt` return the row *kind* would remove the whole class.

Thanks — clean, minimal, well-tested.

PR #158, opened 2026-08-21
fix(slackdesktop): pick the most recently used Slack profile

yml:
Follow-up to #128, targeting that PR's branch rather than `main` so it can be
folded in before the PR lands.

## The problem

I built #128 to use slk with flatpak Slack on Manjaro. The new detection works,
but slk still couldn't authenticate — every workspace failed with
`auth test failed: invalid_auth`.

The cause isn't detection, it's precedence. I have a leftover profile from a
`.deb` install I stopped using in June 2025, sitting next to the flatpak profile
I actually use:

| path | last touched |
| --- | --- |
| `~/.config/Slack` | 2025-06-05 (removed `.deb`) |
| `~/.var/app/com.slack.Slack/config/Slack` | today (flatpak, live) |

`configDirForOS` returns the first candidate that exists, and `~/.config/Slack`
is checked first. So slk read 14-month-dead cookies and tokens:

```
ConfigDir: /home/yml/.config/Slack
  gophers               CONNECT FAIL: auth test failed: invalid_auth
  lincolnloop           CONNECT FAIL: auth test failed: invalid_auth
  ... (all 8 workspaces)
```

Forcing the flatpak dir via `XDG_CONFIG_HOME` made all of them connect, which
confirmed the diagnosis.

Existence alone can't tell the two apart. Both directories hold a full set of
session files — `Cookies`, `storage/root-state.json`, `Local Storage/leveldb`.
Only the timestamps differ.

## The change

Probe each candidate and take the **most recently used** profile instead of the
first one that exists. Candidate order now only breaks ties, so the existing
preferences (xdg > native > flatpak > snap) still hold when timestamps match.

- A directory counts as a profile only when it has **both**
  `storage/root-state.json` and `Cookies`. Either one missing leaves us unable
  to authenticate anyway.
- Its age is the newest mtime among those two plus `Local Storage/leveldb`. A
  running Slack touches leveldb, so it's often the freshest signal.
- `exists` is tracked separately from `live`, so a directory that's present but
  never signed into still gets named in the resulting error rather than falling
  back to a path that was never created.
- The same fix applies on **darwin**, where a leftover unsandboxed profile can
  shadow the App Store one identically. **Windows** is untouched — single
  install location, returns early.

The test seam changed from `exists func(string) bool` to
`probe func(string) profileInfo`. Existing cases were rewritten onto it and kept
(now expressed as tie-breaks), plus new coverage for recency and for
`profileInfoAt` against real temp dirs.

Note that one previous expectation inverted deliberately: the old
`native preferred over flatpak` case is now `native wins ties against flatpak`.
Unconditional native preference is the bug.

## Verification

- The new tests fail against the old first-existing logic and pass with the fix
  (checked by temporarily neutering the recency comparison).
- End to end on my machine, no env override:

```
ConfigDir: /home/yml/.var/app/com.slack.Slack/config/Slack
  gophers / lincolnloop / pysortium / slackyourdata
  smithsoniandigital / washtimes / temporalio   -> all CONNECT OK
```

- Full suite green (53 packages), `go vet` clean, `gofmt` clean.

### DIFF
--- a/cmd/slk/onboarding.go
+++ b/cmd/slk/onboarding.go
@@ -132,6 +132,8 @@ func desktopErrorMessage(err error) string {
 		return "Your system keyring is locked. Unlock it (log in to your desktop session) and retry."
 	case errors.Is(err, slackdesktop.ErrNoSecretService):
 		return "No system keyring/secret service found. slk needs it to read the Slack session."
+	case errors.Is(err, slackdesktop.ErrSecretNotFound):
+		return "No Slack entry found in your keyring. Sign in to the Slack desktop app (and make sure it uses the system keyring), then retry."
 	case errors.Is(err, slackdesktop.ErrDecryptFailed):
 		return "Could not decrypt the Slack session cookie. Please file an issue with your OS + Slack version."
 	default:
--- a/internal/slackdesktop/configdir.go
+++ b/internal/slackdesktop/configdir.go
@@ -21,10 +21,29 @@ func configDirForOS(goos string, getenv func(string) string, exists func(string)
 		}
 		return second
 	default: // linux and others
-		if x := getenv("XDG_CONFIG_DIR"); x != "" {
-			return filepath.Join(x, "Slack")
+		home := getenv("HOME")
+		var candidates []string
+		if x := getenv("XDG_CONFIG_HOME"); x != "" {
+			candidates = append(candidates, filepath.Join(x, "Slack"))
 		}
-		return filepath.Join(getenv("HOME"), ".config", "Slack")
+		if x := getenv("XDG_CONFIG_DIR"); x != "" {
+			candidates = append(candidates, filepath.Join(x, "Slack"))
+		}
+		candidates = append(candidates,
+			filepath.Join(home, ".config", "Slack"),
+			// flatpak (com.slack.Slack)
+			filepath.Join(home, ".var", "app", "com.slack.Slack", "config", "Slack"),
+			// snap
+			filepath.Join(home, "snap", "slack", "current", ".config", "Slack"),
+		)
+		for _, c := range candidates {
+			if exists(c) {
+				return c
+			}
+		}
+		// Nothing found: fall back to the primary location so the
+		// not-found error references the conventional path.
+		return candidates[0]
 	}
 }
 
--- a/internal/slackdesktop/configdir.go
+++ b/internal/slackdesktop/configdir.go
@@ -4,25 +4,82 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"time"
 )
 
+// profileInfo describes a candidate Slack desktop profile directory.
+type profileInfo struct {
+	// lastUsed is the newest mtime among the profile's session files. It
+	// stands in for "when was this Slack install last signed in with", and
+	// is only meaningful when live is true.
+	lastUsed time.Time
+	// live reports whether the directory holds the files needed to pull a
+	// session out of it.
+	live bool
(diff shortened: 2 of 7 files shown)


gammons:
Thanks for this, and thanks especially for the independent flatpak confirmation on #128 — that unblocked a question that had been sitting unanswered for a month.

**First, the reason CI never ran.** It wasn't permissions or a fluke. This PR was based on `fix/linux-packaged-config-dir` (#128's branch), not `main`, and `.github/workflows/ci.yml` filters on `pull_request: branches: [main]`. So no `ci` check suite was ever *created* — `check-suites` for your head SHA returned only GitGuardian. A rebase wouldn't have fixed that; only retargeting would. I've retargeted this to `main` and reopened it to trigger the workflow, and CI is running now for the first time. Until #128 lands, the diff here will show both PRs' commits; it'll shrink to just yours automatically once #128 merges.

Since CI hadn't run, I built and tested it myself, stacked on a rebased #128 on current main:

| check | result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test ./internal/slackdesktop/... -race` | pass (all 6 packaging + 5 recency subtests, both `profileInfoAt` tests) |

**On the substance:** this is the next bug behind #128 and I agree it needs to ship with it. #128 picks by existence order and `~/.config/Slack` is probed before the flatpak path, but a leftover profile from a removed `.deb` is indistinguishable from a live one by existence alone — both have `Cookies`, `storage/root-state.json`, and `Local Storage/leveldb`. Recency is the right discriminator given that. And it degrades cleanly to #128's semantics on ties, since `After` is strict so the earliest candidate wins — good, and the comment saying so is at exactly the right line.

Worth noting this isn't Linux-only: main's own onboarding error text already tells users *"If you have had more than one Slack build installed (App Store and standalone), sign out of the one you no longer use."* That's a documentation workaround for the bug you're fixing in code.

Your tests are the best of the three PRs I looked at in this area — injected `probeFrom` for the path-precedence logic, real `t.TempDir()` fixtures with `os.Chtimes` for `profileInfoAt` (correct: that function's whole job is filesystem interaction, so mocking it would test nothing), plus darwin recency coverage so the cross-platform claim is actually tested. And you called out the deliberate inversion of `native preferred over flatpak` rather than slipping it in.

**One design change before I merge:**

**An explicit `XDG_CONFIG_HOME` should win outright, not just break ties.** Under #128 it's an absolute override; here it competes on recency like any other candidate — your own test codifies this (`xdg config home competes on recency like any native path` asserts a fresh flatpak beats a stale `/cfg/Slack`). That removes the escape hatch your own description relies on ("Forcing the flatpak dir via `XDG_CONFIG_HOME` made all of them connect"). An mtime heuristic should never override a variable the user deliberately set. Return it immediately when it probes live, and flip that test to assert the override wins.

Smaller, non-blocking:

- The `"Cookies"` literal is now duplicated between `profileInfoAt` and `cookie.go`. I checked and it's *not* currently a bug — `cookie.go:110-113`'s `Network/Cookies` branch is Windows-only and Windows returns early from the probing code — but if Slack follows Chromium upstream and moves cookies on Linux, `profileInfoAt` would silently classify every modern profile as dead. Worth a shared helper.
- Add a comment noting mtime is defeated by `cp -r`, rsync without `-a`, and restore-from-backup. It's still the best available signal short of parsing `root-state.json`, just worth flagging.
- `candidates[0]` can't panic today (darwin always 2, linux always >= 3) but it's fragile to a future `case` being added.

Sequencing: #128 merges first (rebase-merge, not squash, so this PR isn't stranded), then this.

PR #160, opened 2026-08-25
fix: fetch the full workspace custom emoji list

rfist:
## Problem

Custom emoji resolution was limited to the conversation restored at startup.

`conversations.view` returns the custom emoji used by that conversation, not the workspace's complete custom emoji set. slk stored that partial map during bootstrap and then skipped the background `emoji.list` request whenever the map was non-empty.

As a result, custom emoji used in other channels rendered as literal `:name:` text in messages and reaction pills. Standard emoji and custom emoji present in the restored channel continued to work, which made the failure appear channel-specific.

## Fix

Treat the map returned by `conversations.view` as an early, conversation-scoped subset:

- Publish it immediately so the restored channel can render without waiting for another request.
- Always fetch `emoji.list` once the workspace connects.
- Replace the subset with the authoritative workspace-wide map when that request succeeds.
- Preserve the bootstrap subset if `emoji.list` fails instead of clearing working emoji data.

Restoring the background write means it can now overlap with workspace-switch reads. `WorkspaceContext` therefore publishes the map through an `atomic.Pointer`, following the existing user-group map pattern.

The `conversations.view` documentation is also corrected to state that its emoji data is conversation-scoped and does not replace `emoji.list`.

The request cost is one `emoji.list` call per workspace connection.

## Tests

New tests cover:

- The zero-value `WorkspaceContext` returning a non-nil empty map.
- The workspace-wide result replacing rather than merging with the bootstrap subset.
- Concurrent publication and reads under the race detector.

Verification:

- `go test -race ./cmd/slk`
- `go test ./...`
- `go vet ./...`
- `go build ./cmd/slk`
- Verified in a live workspace: after starting on one channel, custom emoji unique to another channel rendered correctly in message text and reaction pills.

### DIFF
--- a/cmd/slk/customemoji.go
+++ b/cmd/slk/customemoji.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"context"
+
+	"github.com/gammons/slk/internal/ui"
+)
+
+// customEmojiLister is the one-method slice of *slackclient.Client that
+// the emoji fetch needs. Mirrors reconnectClient in reconnect_sync.go:
+// a narrow interface keeps the fetch testable without a live client,
+// and fails compilation if the surface grows.
+type customEmojiLister interface {
+	// ListCustomEmoji is emoji.list: the workspace's entire custom
+	// emoji set. Distinct from conversations.view's Emojis field,
+	// which is only the emoji the fetched conversation happens to use.
+	ListCustomEmoji(ctx context.Context) (map[string]string, error)
+}
+
+// fetchWorkspaceEmoji publishes the workspace's full custom emoji set
+// and tells the UI to re-render with it.
+//
+// It runs UNCONDITIONALLY, and that is the whole point of it. Bootstrap
+// may already have published a map, but conversations.view returns only
+// the emoji the restored conversation uses — a per-conversation subset,
+// not the workspace's set. An earlier version skipped emoji.list
+// whenever that subset was non-empty, which left every channel other
+// than the restored one rendering custom emoji as literal `:name:`.
+// TestFetchWorkspaceEmoji_RunsEvenWhenBootstrapPublishedASubset pins
+// that so the short-circuit cannot come back.
+//
+// Best-effort: on error nothing is published and nothing is sent, so
+// the bootstrap subset (or the built-ins) stays in place rather than
+// being cleared.
+//
+// Intended to be run in a goroutine, after WorkspaceReadyMsg, so it
+// never blocks first paint.
+func fetchWorkspaceEmoji(ctx context.Context, wctx *WorkspaceContext, client customEmojiLister, sender teaSender, teamID string) {
+	emojis, err := client.ListCustomEmoji(ctx)
+	if err != nil {
+		return
+	}
+	wctx.SetCustomEmoji(emojis)
+	if sender != nil {
+		sender.Send(ui.CustomEmojisLoadedMsg{
+			TeamID:      teamID,
+			CustomEmoji: emojis,
+		})
+	}
+}
--- a/cmd/slk/customemoji_test.go
+++ b/cmd/slk/customemoji_test.go
@@ -1,8 +1,13 @@
 package main
 
 import (
+	"context"
+	"errors"
 	"sync"
 	"testing"
+
+	tea "charm.land/bubbletea/v2"
+	"github.com/gammons/slk/internal/ui"
 )
 
 func TestWorkspaceContextCustomEmojiDefaultsEmpty(t *testing.T) {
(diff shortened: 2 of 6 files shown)


gammons:
Confirmed, and the bug is worse than "missing feature" — it's a regression I introduced. `cmd/slk/main.go:2118` short-circuits the `emoji.list` fetch whenever `len(wctx.CustomEmoji) > 0`, and `main.go:2433` populates that map from `res.Emojis`, which is `conversations.view`'s *conversation-scoped* emoji set (`internal/slack/boot/view.go:333`). So on the normal boot path the workspace-wide fetch is skipped and every channel except the restored one renders custom emoji as literal `:name:`. The misleading comment at `boot/view.go:328-332` ("Emojis replaces the emoji.list call") is what caused it, and your correction to it is valuable — that comment should warn future readers, not mislead them.

Two things you got right that a careless fix would have missed:

- **The `atomic.Pointer` conversion is genuinely required, not gold-plating.** On main the background write at `main.go:2125` was gated behind the `len(...)>0` check so it almost never fired, which is what the exemption comment at `main.go:195-198` was leaning on. Removing the gate makes that write unconditional and puts it in a live race with the workspace-switch read at `main.go:1911`. Following the `userGroups` pattern verbatim is the right call.
- **Failure semantics.** `emoji.list` failure returns early without publishing, so the bootstrap subset survives instead of being cleared. Correct.

Cost analysis holds up. The goroutine sits inside the per-workspace connect block and fires *after* `WorkspaceReadyMsg`, so cold-start paint isn't blocked — exactly one extra request per workspace per launch, not per reconnect. `emoji.list` is Tier 2 (20/min), no rate-limit risk. I checked slack-go v0.23.0's `GetEmojiContext` too: single unpaginated POST, and Slack returns the full set when no `limit` is passed, so there's no pagination gap. Downstream propagation is right — `SetCustomEmoji` → `SetEmojiCustoms` (`messages/model.go:1444`) nils the render cache, so already-rendered messages repaint.

**One change before I merge: the actual bug has no test coverage.**

All three new tests cover the `WorkspaceContext` accessor — nil-safety, replace-not-merge, `-race`. They pass, they're fine, but **re-adding `if len(wctx.CustomEmoji()) > 0 { return }` would leave the whole suite green.** The tests guard the refactor, not the regression.

The fetch is inline in `run()`, which is why it isn't testable. Extract it to a named `fetchWorkspaceEmoji(ctx, wctx, p, teamID)` taking a small client interface, then assert that a non-empty bootstrap map does *not* suppress the `emoji.list` call.

Two smaller things:

- `ListCustomEmoji(ctx)` (`internal/slack/client.go:834-835`) discards `ctx` and calls the non-context `GetEmoji()`. Pre-existing, but this PR promotes it from "rarely runs" to "always runs", so a shutdown can no longer cancel it. One line: `c.api.GetEmojiContext(ctx)`.
- No disk cache, so every cold start refetches. Not a blocker since it's background and post-paint, but given how much we care about cold start I'd like a SQLite-backed cache with a TTL as a follow-up. Happy to file that separately.

And to be explicit since it's easy to misread: "custom emoji management" being a non-goal refers to *managing* emoji. Rendering them is core, and half-broken rendering is worse than none.

Add the regression test and I'll merge.

gammons:
@rfist I pushed two commits to this branch rather than leave it sitting — you'd enabled maintainer edits, so I took you up on it. Nothing of yours was rewritten or force-pushed; your commit is untouched and you're co-authored on mine. Shout if you'd rather I'd waited.

Your diagnosis was right and the `atomic.Pointer` conversion is genuinely required, not gold-plating — I verified that the write at `main.go:2125` really does become unconditional and race the workspace-switch read at `main.go:1911`, so the old exemption comment no longer held. Good catch. The `boot/view.go` doc correction is the most valuable part of the whole change: that comment is what caused the bug, and now it warns instead.

**1. The regression had no test.**

All three of your tests covered the `CustomEmoji()`/`SetCustomEmoji()` accessor — nil-safety, replace-not-merge, `-race`. They're fine, but re-adding

```go
if len(wctx.CustomEmoji()) > 0 { return }
```

left the whole suite green. That's the exact short-circuit that caused the bug, so the one thing that could actually regress was unguarded.

The reason it wasn't testable is that the fetch was an anonymous goroutine inline in `run()`. I extracted it to `fetchWorkspaceEmoji` behind a one-method `customEmojiLister` interface, following the `reconnectClient` / `teaSender` pattern already sitting in `reconnect_sync.go:34`. Four tests now cover it:

- `RunsEvenWhenBootstrapPublishedASubset` — the guard. I verified it by reintroducing the short-circuit; it fails with `ListCustomEmoji calls = 0, want 1`.
- `ErrorKeepsBootstrapSubset` — a failed fetch must not clear the partial set or tell the UI to repaint.
- `SendsLoadedMsgForItsTeam` — the follow-up carries its `TeamID`, so it can't repaint another workspace.
- `NilSenderDoesNotPanic` — it runs in a bare goroutine, so a panic there takes the process down.

**2. `ListCustomEmoji` was dropping its context.**

Not something you introduced, but your change is what makes it matter:

```go
func (c *Client) ListCustomEmoji(ctx context.Context) (map[string]string, error) {
	emojis, err := c.api.GetEmoji()   // ctx never used
```

slack-go's `GetEmoji` hardcodes `context.Background()` internally. Unused function parameters draw no compiler or `go vet` complaint, so nothing flagged it. Harmless while this only fired on the rare `conversations.history` fallback — but it now runs on every workspace connect, so a shutdown couldn't cancel it. Switched to `GetEmojiContext` and moved the `SlackAPI` interface entry to the Context variant, which also lines it up with `GetUserGroupsContext` / `GetBotInfoContext` / `GetUsersInConversationContext` around it.

I also merged current `main` in, since the branch predated the tree-wide gofmt pass and `gofmt` is now an enforced lint check.

Green across the board: `go build`, `go vet`, `gofmt -l` clean, `go test ./... -race` 54 packages, and CI is passing.

**On the cold-start cost** — I checked your reasoning and it holds. The goroutine fires after `WorkspaceReadyMsg`, so first paint isn't blocked; it's one extra request per workspace per launch, not per reconnect; `emoji.list` is Tier 2 (20/min) so there's no rate-limit exposure; and slack-go issues a single unpaginated POST that returns the full set, so there's no pagination gap.

The one thing I'd still like, as a **separate PR** rather than here: the map isn't persisted, so every cold start refetches it. A SQLite-backed cache with a TTL would make cold starts request-free. Happy to file it if you don't want it.

Worth knowing this also just got more valuable: #174 landed an ASCII fast path in `text.Fold` that cut emoji-picker filtering from ~135 ms to ~1.6 ms per keystroke on a large catalog. That was latent until this PR started loading the full workspace list — so the two belong together, and this is the one that switches it on.

Merging.

rfist:
@gammons No objection at all — maintainer edits were on precisely so you wouldn't have to sit and wait on my timezone, and co-authorship is more than fair. Pushing beats blocking.

The test criticism was right, and it's the useful kind of right. My three tests asserted that the accessor I had just introduced does what I said it does, which is close to a tautology — none of them fail if the short-circuit comes back, and the short-circuit is the only thing that can actually regress. "The tests guard the refactor, not the regression" is a sentence I'd like to keep.

The seam placement is better than what I'd have reached for. I would have put the fake at the `SlackAPI` boundary and grown that interface; a one-method `customEmojiLister` next to `reconnectClient` / `teaSender` in `reconnect_sync.go` keeps the double local to the test and matches a pattern already in the file. `NilSenderDoesNotPanic` is the one I wouldn't have thought to write, and it's the one that matters most in a bare goroutine — a panic there is not a failed fetch, it's a dead process.

On `ListCustomEmoji` dropping its context: agreed, and your explanation of why nothing caught it is the part worth writing down. An unused parameter on a method that exists to satisfy an interface is invisible to both the compiler and `vet` — the signature is load-bearing for the interface, so nothing has grounds to complain that the body ignores it. Moving the `SlackAPI` entry to the Context variant alongside `GetUserGroupsContext` and friends is the right shape; a lone non-Context method in that list was a smell nobody had reason to look at while it barely ran.

**I'd like to take the SQLite cache follow-up** rather than have you file it. Questions I'd want to settle in the issue before writing any of it:

- **TTL.** Custom emoji churn is workspace-culture dependent — some workspaces add a few a year, some a few a day. I'd start at 24h and treat that as tunable rather than a constant, but I don't have a principled number.
- **Invalidation.** slk doesn't subscribe to `emoji_changed` today (no handler anywhere in the tree). Handling it would make the TTL a backstop rather than the mechanism, at the cost of a new event path. Probably worth it, but it's the decision that sizes the change.
- **Precedence at boot.** Cache-then-revalidate keeps cold start request-free but can paint one frame of stale `:name:` for an emoji added since; fetch-first throws away most of the benefit. I'd default to serve-cache-then-refresh-in-background, reusing this PR's repaint path, since `SetCustomEmoji` → `SetEmojiCustoms` already nils the render cache.
- **Shape.** `frecent_emoji` is already in `db.go`, so there's precedent for emoji living in the cache DB; a `custom_emoji(team_id, name, url, fetched_at)` table keyed per workspace looks like the obvious fit rather than a blob.
- **Alias targets.** The map stores `alias:target` alongside URLs. Round-tripping those through SQLite is trivial but worth an explicit test, because a broken alias degrades to a literal `:name:` — the exact symptom this PR just fixed.

I'll open the issue with that and pick it up after the two PRs I still have in flight.

Good spot on #174. I hadn't connected them: the ASCII fast path in `text.Fold` was latent while only a per-conversation handful of emoji ever reached the picker, and this change is what turns a large workspace catalog into the common case. Ordering worked out — the other way round would have shipped a visible input stall.

PR #161, opened 2026-08-25
fix(kitty): encode images at the terminal's measured cell size

rfist:
## Problem

The kitty renderer encoded image payloads using a hardcoded 8×16 pixels per terminal cell:

```go
pxW := target.X * 8
pxH := target.Y * 16
```

The terminal still placed the image correctly because the kitty graphics protocol uses `c=<columns>` and `r=<rows>` to define its cell footprint. However, it had to upscale the undersized PNG across that footprint, reducing image quality on HiDPI terminals.

For example, a terminal reporting 14×34 pixel cells received only 57% of the available horizontal resolution and 47% of the available vertical resolution. The result was visibly soft, especially for screenshots containing small text.

slk already measured the actual terminal cell geometry through `CellPixels`, but those measurements were only used by the sixel encoder.

## Fix

Use the measured terminal cell size when building kitty PNG payloads:

```go
cw, ch := measuredCellPixels()
pxW := target.X * cw
pxH := target.Y * ch
```

The existing sixel-specific helper is renamed to `measuredCellPixels` because both encoders now use it:

- Sixel needs the measurements for correct on-screen dimensions.
- Kitty needs them to transmit enough source pixels for the terminal's native cell resolution.

The kitty placement dimensions remain unchanged, so this affects image resolution without changing layout. Terminals that do not provide usable measurements retain the previous 8×16 fallback.

On Retina-sized cells this produces larger payloads, roughly 3× in the observed geometry. The cell measurement is fixed before rendering starts, and encoded payloads remain cached per image and target cell size, so the resize and PNG encoding cost is paid once per cached payload.

## Tests

New kitty payload tests decode the actual cached PNG and verify:

- Its dimensions match the measured terminal cell size.
- The previous 8×16 behavior remains as the fallback.
- HiDPI measurements produce a higher-resolution payload than the old hardcode.

The existing sixel cell-geometry tests continue to cover measured dimensions, invalid measurements, and raster scaling.

Verification:

- `go test -count=1 -race ./internal/image`
- `go test -count=1 ./...`
- `go vet ./...`
- `go build ./cmd/slk`
- `git diff --check`
- Manual visual test passed with the branch build.

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -108,7 +108,7 @@ type WorkspaceContext struct {
 	// Edge is the edgeapi client for this workspace: the
 	// conditional-revalidation and server-side-search endpoints. Nil
 	// only if construction failed, and every caller nil-checks.
-	Edge       *edge.Client
+	Edge *edge.Client
 	// EdgeHealth records whether edge resolution is working for this
 	// workspace this session. bootstrap marks it degraded on a
 	// wholesale failure; the user resolver reads it to skip batch
@@ -1063,14 +1063,13 @@ func run() error {
 	// per visible avatar per redraw would dominate the bandwidth budget.
 	avatarCache := avatar.NewCache(imageFetcher, imgpkg.KittyRendererInstance(), proto == imgpkg.ProtoKitty)
 
-	// Cell pixel metrics for sizing decisions.
+	// Cell pixel metrics for image encoding. Sixel uses them for its
+	// absolute raster dimensions; kitty places by cell but uses them to
+	// transmit enough source pixels for the terminal's native cell
+	// resolution. SetCellPixels retains its 8x16 fallback when the
+	// terminal reports no usable geometry.
 	pxW, pxH := imgpkg.CellPixels(int(os.Stdout.Fd()))
 	debuglog.ImgRender("cell pixels: %dx%d", pxW, pxH)
-	// Sixel encodes at absolute pixel dimensions — the terminal paints
-	// one sixel pixel per device pixel rather than scaling into a cell
-	// box the way kitty does. Hand it the measured metrics so an image
-	// occupying N rows of layout is encoded N*cellHeight pixels tall and
-	// actually lands on those rows.
 	imgpkg.SetCellPixels(pxW, pxH)
 
 	// Wire the inline-image pipeline into the messages pane. SendMsg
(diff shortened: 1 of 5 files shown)


trevor-vaughan:
:+1: Fix works well for me

gammons:
Reviewed. Merging this.

The diagnosis is right and it's the strongest kind of bug — an inconsistency where one stage disagrees with the rest of its own pipeline. Every other stage already uses the *measured* cell size:

- `cmd/slk/main.go:1067` measures it, `main.go:1086` puts it in `ImageContext.CellPixels`
- `internal/ui/imgrender/imgrender.go:324` computes `pixelTarget` from it
- `imgrender.go:329` fetches a thumb at least that large
- `internal/image/fetcher.go:473-477` downscales to exactly `pixelTarget`

...and then `internal/image/kitty.go:229-230` threw it away and resized down to a hardcoded 8x16. On a 14x34 cell that discards ~73% of pixels we'd already paid to fetch and decode. Self-inflicted, and it's worst on exactly the terminals with the best kitty support.

I checked the fallback path too: `measuredCellPixels()` returns 8/16 when unset and `SetCellPixels` rejects non-positive input, so terminals that report nothing render byte-identically to before. No sixel regression — that diff is a rename plus a comment. And iTerm2 falls back to half-block per `wiki/Terminal-Compatibility.md:13`, so it never enters this path at all.

The test decodes the actual cached PNG and asserts its dimensions rather than re-deriving the arithmetic, which is the right way to test this.

Thanks also @trevor-vaughan for confirming it visually.

Follow-ups I'll file separately, none blocking:

- The comment at `kitty.go:128-131` ("~10-30KB per image, ~20MB for 1000 images") is now stale. Measured on a 360x240 source at a 40x12 target: 8x16 gives ~230KB, 14x34 gives ~822KB, 20x40 gives ~1.27MB. That's a 3.6-5.5x jump and `k.payloads` still has no eviction. The comment already anticipates an LRU; this is the moment to add it. Mitigating factor is that `fetcher.decoded` was *already* holding the larger decoded images, so this aligns the payload cache with memory we were spending anyway rather than introducing a new class.
- Same multiplier applies to terminal write bandwidth on every channel switch with images, which is felt over ssh/mosh. Worth a line in `wiki/Terminal-Compatibility.md`.
- `payloadKey := phKey` is `{id, cellsX, cellsY}` and doesn't include cell pixel size. Safe today because `SetCellPixels` is called once at `main.go:1074` before any render, but if we ever re-measure on SIGWINCH or a DPI change, cached payloads go silently stale. Worth adding `cw, ch` to the key or a comment pinning the one-shot invariant.

PR #162, opened 2026-08-25
fix: add newly opened conversations to the channel finder

rfist:
## Problem

When somebody messages you for the first time while slk is running, the new conversation appears in the sidebar and you can open it and reply normally.

But if you later press Ctrl+P and search for that person, they do not appear in the channel picker. The conversation exists and is usable from the sidebar—the picker is simply out of date. It only picks up the new DM after switching workspaces or restarting slk.

The same problem affects newly opened group DMs and channels.

Internally, Slack delivers these conversations through `im_created`, `im_open`, `mpim_open`, `group_joined`, and `channel_joined` WebSocket events. `OnConversationOpened` already persisted the new conversation and updated the active sidebar, but the running Ctrl+P finder never received the new item.

## Fix

Carry the finder item in `ConversationOpenedMsg` and upsert it into the active workspace's channel finder alongside the sidebar item.

The new `channelfinder.Model.Upsert` behavior:

- Adds conversations that are not yet present.
- Replaces an existing entry with the same channel ID instead of creating a duplicate.
- Converts an existing browseable result into a joined conversation.
- Preserves the existing `LastVisited` value when the incoming event has no recency information.
- Refilters immediately when Ctrl+P is already open.

Inactive-workspace events continue to update `WorkspaceContext.Channels` and `WorkspaceContext.FinderItems`. Switching to that workspace later rebuilds both UI models from the updated context.

## Tests

New in this PR:

- Finder insertion and visible-query refiltering (`TestUpsertRefiltersVisibleQuery`).
- Same-ID replacement without duplication (`TestUpsertReplacesByIDAndPreservesLastVisited`).
- `LastVisited` preservation during replacement.
- Converting an existing browseable (non-joined) entry into a joined one in place, rather than duplicating it (`TestUpsertConvertsBrowseableEntryInPlace`).
- Active-workspace `ConversationOpenedMsg` reaching both the sidebar and channel finder, and inactive-workspace messages leaving the active finder unchanged.

The `im_created`, `im_open`, `mpim_open`, `group_joined`, and `channel_joined` dispatch tests already existed on main and are unchanged here; they're listed for context, not as new coverage.

Verification:

- `go test -count=1 -race ./internal/ui/channelfinder ./internal/ui ./cmd/slk`
- `go test -count=1 ./...`
- `go vet ./...`
- `go build ./cmd/slk`
- `git diff --check`

Manual first-time-DM verification was not available because it requires another Slack user to initiate a previously nonexistent conversation.

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -4478,7 +4478,7 @@ func (h *rtmEventHandler) OnThreadSubscriptionChanged(channelID, threadTS, lastR
 // unread/last-read state), upserts the SQLite cache row, mirrors
 // channelNames/Types maps used by the notifier, and — if the
 // workspace is active — forwards a ConversationOpenedMsg to the UI
-// so the live sidebar updates.
+// so the live sidebar and channel finder (Ctrl+P) both update.
 func (h *rtmEventHandler) OnConversationOpened(ch slack.Channel) {
 	if h.wsCtx == nil {
 		return
@@ -4532,8 +4532,9 @@ func (h *rtmEventHandler) OnConversationOpened(ch slack.Channel) {
 		return
 	}
 	h.program.Send(ui.ConversationOpenedMsg{
-		TeamID: h.workspaceID,
-		Item:   item,
+		TeamID:     h.workspaceID,
+		Item:       item,
+		FinderItem: finderItem,
 	})
 }
 
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -176,7 +176,7 @@ type WorkspaceContext struct {
 	// decide whether to draw the "Threads list unavailable" banner.
 	SubscriptionsAvailable bool
 	Channels               []sidebar.ChannelItem
-	// FinderItems is the list shown in the Ctrl+T finder: the channels
+	// FinderItems is the list shown in the Ctrl+P finder: the channels
 	// the user has joined. Channels they have not joined are not held
 	// here at all — they arrive per query from the finder's debounced
 	// channels/search and live only in the finder component.
@@ -4509,7 +4509,7 @@ func (h *rtmEventHandler) OnConversationOpened(ch slack.Channel) {
 		// path. On dedupe, the existing finder entry was added at
 		// bootstrap (or a prior open) and carries no unread state to
 		// refresh, so re-appending would double-list the channel in
-		// Ctrl+T.
+		// Ctrl+P.
 		finderItem.LastVisited = h.wsCtx.LastVisitedByChannel[ch.ID]
 		h.wsCtx.FinderItems = append(h.wsCtx.FinderItems, finderItem)
 	}
--- a/internal/ui/app_test.go
+++ b/internal/ui/app_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/gammons/slk/internal/cache"
 	"github.com/gammons/slk/internal/ids"
 	imgpkg "github.com/gammons/slk/internal/image"
+	"github.com/gammons/slk/internal/ui/channelfinder"
 	"github.com/gammons/slk/internal/ui/compose"
 	"github.com/gammons/slk/internal/ui/messages"
 	"github.com/gammons/slk/internal/ui/sidebar"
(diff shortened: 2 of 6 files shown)


gammons:
Verified and this is the right fix in the right layer. `cmd/slk/main.go:4520-4523` sends `ConversationOpenedMsg` with only the sidebar item, and `internal/ui/reducer_workspace.go:74-77` calls `UpsertItem` and nothing else — `wsCtx.FinderItems` gets updated at `main.go:4500` but that only reaches the UI on the next `SetItems`, i.e. workspace activation. So a first-time DM is usable from the sidebar and invisible in Ctrl+P until restart. Exactly as you describe.

I specifically checked the case that usually breaks this kind of fix, and you got it right: browseable (non-joined) search results *do* live in `m.items` — `SetBrowseable` appends them there with `Joined=false` (`internal/ui/channelfinder/model.go:160-166`) — so `Upsert`'s single-slice ID scan converts rather than duplicating. The reverse holds too: a later `SetBrowseable` keeps `Joined||Synthetic` entries and skips dups via the `have` set. `Synthetic` is preserved on replace, and the `LastVisited == 0` preservation is correct given `main.go:4499` only sets it on the not-replaced branch.

The Ctrl+T to Ctrl+P comment corrections are a genuine fix — those comments were just wrong.

**Two small things before I merge:**

1. **Add the browseable-to-joined test.** The description lists "converts an existing browseable result into a joined conversation" as covered, but there's no test for it in the diff — and it's the dedupe path most likely to regress later, since it depends on the non-obvious fact that browseable items share `m.items`. Roughly: `SetBrowseable([{ID:"C9", Joined:false}])`, then `Upsert({ID:"C9", Joined:true})`, assert one item with `Joined == true`. (The description also credits the five WS-event dispatch tests to this PR, but those look pre-existing.)

2. **Guard the empty item** at `reducer_workspace.go:76` — `if m.FinderItem.ID != ""`. `ConversationOpenedMsg` is exported with one sender today, but a future sender that forgets the new field would silently append an `ID: ""` row.

One thing to note, not asking you to fix it: on the `replaced` branch (`main.go:4484-4491`) `wsCtx.FinderItems` is deliberately not updated, but the message with the fresh `finderItem` is still sent. So the live finder now shows the updated name while `FinderItems` holds the stale one — switch away and back and it reverts. Pre-existing divergence that your change makes observable. I'll file it.

Also drop the `Edge *edge.Client` gofmt hunk — it also appears in #161 and they'll collide. I'm enforcing gofmt in CI in #171, which removes that whole category.

rfist:
Thanks for the review — addressed both.

- Added `TestUpsertConvertsBrowseableEntryInPlace`: seeds a browseable (non-joined) `C9` via `SetBrowseable`, then `Upsert`s it as joined, and asserts one item with `Joined == true`. Confirmed it fails against the old append-only `Upsert` before restoring the real one.
- `reducer_workspace.go` now skips the finder upsert when `FinderItem.ID` is empty, with a comment on why — defensive against a future `ConversationOpenedMsg` sender that forgets to set it.
- Rebased onto main; the `Edge *edge.Client` gofmt hunk is gone on its own now that #171 reformatted the tree, so there's nothing left to revert there.
- Corrected the PR description's test list — it credited five WS-dispatch tests that already existed on main; removed that claim and added the new dedupe test.

Locally: `gofmt -l` clean, build, vet, and `go test ./...` green, plus `-race` on `channelfinder` and `ui`.

rfist:
@gammons, friendly reminder on this one 🙂 Can we have a "ready for merge" label here as well?

gammons:
@rfist Apologies for the wait — yes, this is ready.

I verified both items landed rather than taking the comment on faith:

- `TestUpsertConvertsBrowseableEntryInPlace` covers the browseable→joined conversion. I confirmed it has teeth by reverting `Upsert` to append-only — the test fails, and passes with the real implementation. That's the dedupe path that was previously untested.
- The empty-ID guard at `reducer_workspace.go:81` is there with the comment explaining the defensive intent.

Merges cleanly against current main, gofmt clean, 54 packages green under `-race`.

And thank you for correcting your own description when you realized it credited pre-existing tests — that kind of self-honesty about test coverage is exactly what makes these easy to review.

Labeling it ready to merge.

PR #163, opened 2026-08-26
feat(ui): add y keybinding to copy message text

Yukaii:
## Summary
- Adds `CopyMessage` keybinding on `y` in normal mode for copying the selected message or thread reply text.
- Reconstructs mrkdwn for Block Kit rich text blocks via `MessageTextSource`.
- Writes text to system clipboard via OSC 52 and displays a `Copied N chars` status bar toast.
- Includes unit tests in `internal/ui/copy_message_test.go` and updates documentation.

### DIFF
--- a/docs/STATUS.md
+++ b/docs/STATUS.md
@@ -53,7 +53,7 @@ Last updated: 2026-05-03
 - [x] Message editing (`E` on own message; reuses compose with stash/restore draft)
 - [x] Message deletion (`D` on own message; centered confirmation overlay)
 - [x] Paste-to-upload via `Ctrl+V` in insert mode (clipboard image, file path, or text fallback) using Slack's V2 file-upload API; multiple attachments + caption send together; status-bar progress + error toasts
-- [x] OSC 52 clipboard integration for message selection and permalink copying
+- [x] OSC 52 clipboard integration for message selection, message text (`y`), and permalink copying (`Y` / `C`)
 - [x] In-place update on `message_changed` echoes (no duplicate row on edit)
 - [x] Live removal on `message_deleted` echoes from any client
 - [x] @mention autocomplete in compose (inline picker, translates to <@UserID> on send)
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -1008,6 +1008,38 @@ func (a *App) toggleReactionOnMessageItem(channelIDStr string, msg messages.Mess
 	}
 }
 
+// copyMessageOfSelected copies the text of the currently-selected message or
+// thread reply to the system clipboard via OSC 52 and emits a status-bar toast.
+func (a *App) copyMessageOfSelected() tea.Cmd {
+	var msg messages.MessageItem
+	switch a.focusedPanel {
+	case PanelMessages:
+		m, ok := a.messagepane.SelectedMessage()
+		if !ok {
+			return nil
+		}
+		msg = m
+	case PanelThread:
+		reply := a.threadPanel.SelectedReply()
+		if reply == nil {
+			return nil
+		}
+		msg = *reply
+	default:
+		return nil
+	}
+
+	text := messages.MessageTextSource(msg)
+	if text == "" {
+		return func() tea.Msg { return ToastMsg{Text: "Message has no text"} }
+	}
+	n := len([]rune(text))
+	return tea.Batch(
+		a.clipboardWrite(text),
+		func() tea.Msg { return statusbar.CopiedMsg{N: n} },
+	)
+}
+
 // copyPermalinkOfSelected resolves the currently-selected message or thread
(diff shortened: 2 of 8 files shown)


gammons:
Thanks — this is well-built and I want it. Docs updated in three places, roadmap line removed, tests above the repo's bar. That's the discipline I'm looking for.

**First: the red CI is not your fault.** The failure is `TestBackgroundFetchRetriesAfterBackoffExpiry` in `internal/slack/membership`, which isn't in your 8-file diff. It's a flaky test on main — `EnsureFresh` pushes synchronously, so the test's `waitForPush` barrier was being satisfied by its own call rather than by the background fetch. I've fixed it in #171. Rebase once that lands and this should go green. I built your branch locally and all of `./internal/ui/...` passes.

**One change I want: drop the `c` alias.**

`c` is *change* in vim, and slk already has that operation bound to `E` (edit message). Binding `c` to copy in an app whose whole selling point is vim muscle memory teaches the wrong reflex, and having `c` and `E` occupy the same conceptual slot is incoherent. It also burns the last plausible operator-prefix key in a single-letter namespace that's already ~30 deep — once `c` ships as an immediate command, reclaiming it for a `c{motion}` chord is a breaking change.

`y` alone is right, and `y`/`Y` (message text / permalink) is already a clean pair. Four keys for two actions is the opposite of the "tiny" thing we're going for.

Concretely:
- `internal/ui/keys.go:103` → `key.NewBinding(key.WithKeys("y"), key.WithHelp("y", "copy message"))`
- Update `TestDefaultKeyMap_CopyMessage` (it currently asserts `Help().Key == "y/c"`, which welds the UX decision into a test) and drop `TestCopyMessage_FromMessagesPane_C`
- Update `wiki/Keybindings.md`, `wiki/Features.md`, `docs/STATUS.md` to say `y` only

Also worth fixing while you're in there: the `mode_normal.go:13` header comment says "message ops: y (copy permalink)" — already wrong on main (it's `Y`/`C`).

On single `y` vs the `yy` the roadmap specified: your call is the better one. `yy` would need a pending-key timeout and there's no `y`-prefixed operator namespace to disambiguate against. I'm fine with the roadmap being wrong here.

Two notes, neither blocking:
- No `ViewThreads` guard on `PanelMessages`, so in the Threads view the messages pane holds the threads list. `openLinksOfSelected` and `copyPermalinkOfSelected` have the identical gap, so you're consistent with existing precedent — I'll fix all three together later.
- `wiki/Tradeoffs-and-Non-Goals.md:17` lists "in-app toasts" as not planned, which is stale relative to the codebase (`app.go:1089` already does this). Your `ToastMsg` follows established precedent; the doc is what's wrong.

Drop `c`, rebase after #171, and I'll merge.

gammons:
#171 has landed, so the `membership` test flake I mentioned is gone. I re-ran CI here and `test` is now green.

One new thing surfaced, though: #171 also turned on `gofmt` as an enforced lint check, and it caught a formatting issue CI previously couldn't see:

```
internal/ui/copy_message_test.go:7:1: File is not properly formatted (gofmt)
```

So the updated ask is:

1. `gofmt -w internal/ui/copy_message_test.go`
2. Drop the `c` alias (see my review above) — `keys.go:103`, the `y/c` help assertion in `TestDefaultKeyMap_CopyMessage`, `TestCopyMessage_FromMessagesPane_C`, and the three doc files
3. Rebase onto main

Then I'll merge.

Yukaii:
Thanks for the review! I've addressed all the feedback:
1. Rebased onto `main` (incorporating the fix from #171).
2. Dropped the `c` alias so only `y` copies message text, updated `keys.go`, tests, and docs (`Keybindings.md`, `Features.md`, `STATUS.md`).
3. Fixed the header comment in `mode_normal.go`.
4. Formatted `internal/ui/copy_message_test.go` with `gofmt`.

gammons:
Verified — this is good to go.

Checked the `c` removal is complete rather than partial: it's gone from the binding, `TestDefaultKeyMap_CopyMessage` now asserts `"y"`, `TestCopyMessage_FromMessagesPane_C` is renamed to `_Y`, and all three doc files are updated. `CopyPermalink` still holds `Y`/`C` untouched, so nothing regressed there, and `c` is now genuinely free for a future `c{motion}` chord.

The `mode_normal.go` header comment reads correctly now too:

```
//   - message ops: y (copy message), Y/C (copy permalink), E (edit), D (delete),
```

gofmt clean, build/vet green, and `go test ./internal/ui/... -race` passes across all 29 packages locally. CI green.

Thanks for taking the `c` point — I know it's an annoying thing to be asked to remove after you'd already written and documented it, but `c` is *change* in vim and `E` already owns that slot here, so it would have been a reflex we'd be stuck supporting.

Merging.

PR #166, opened 2026-08-31
fix(config): merge leftover [workspaces.<teamID>] blocks

agustif:
## Why

`saveWorkspaceVersionTS` (and theme/width saves) looked only for `[workspaces.<tomlKey>]`. When `tomlKey` fell back to the raw Slack team ID, slk appended a second `[workspaces.T…]` block next to the slug-keyed one from `--add-workspace`. `Load` then refused to start with `workspaces "…" and "…" both reference team_id`.

## What

- Treat a leftover `[workspaces.<teamID>]` next to a slug-keyed block for the same team as prefs, not a duplicate workspace.
- Merge unset fields (`version_ts`, `theme`, `sidebar_width`, …) into the slug block and drop the leftover key.
- Shared `findWorkspaceSectionStart` writes subsequent saves into the existing slug section (or a block whose `team_id` matches).

Two slug-keyed blocks for the same `team_id` are still an error.

## Test plan

- [x] `go test ./internal/config/ ./cmd/slk/`
- [ ] Load a config that has both `[workspaces.my-team]` and `[workspaces.T…]` for the same team; slk should start and keep the slug's theme plus the leftover `version_ts`.

Merge first in this stack (1/4). Independent of the later sidebar/Activity PRs except that they sit on top of this commit.

### DIFF
--- a/cmd/slk/save_theme.go
+++ b/cmd/slk/save_theme.go
@@ -125,6 +125,55 @@ func sanitizeComment(s string) string {
 	return strings.TrimSpace(b.String())
 }
 
+// findWorkspaceSectionStart returns the line index of the [workspaces.*]
+// header to write into for tomlKey/teamID, or -1 if none exists.
+//
+// Order: the literal tomlKey header, then a legacy [workspaces.<teamID>]
+// header, then any slug-keyed block whose team_id field matches. That
+// last fallback is load-bearing — saveWorkspaceVersionTS used to look
+// only for [workspaces.<tomlKey>], and when tomlKey was the raw team ID
+// it appended a second block next to the slug-keyed one, which Load
+// then rejected as a duplicate workspace.
+func findWorkspaceSectionStart(lines []string, tomlKey, teamID string) int {
+	headers := make([]string, 0, 2)
+	if tomlKey != "" {
+		headers = append(headers, "[workspaces."+tomlKey+"]")
+	}
+	if teamID != "" && teamID != tomlKey {
+		headers = append(headers, "[workspaces."+teamID+"]")
+	}
+	for i, line := range lines {
+		t := strings.TrimSpace(line)
+		for _, h := range headers {
+			if t == h {
+				return i
+			}
+		}
+	}
+	if teamID == "" {
+		return -1
+	}
+	sectionStart := -1
+	for i, line := range lines {
+		t := strings.TrimSpace(line)
+		if strings.HasPrefix(t, "[") {
+			if strings.HasPrefix(t, "[workspaces.") {
+				sectionStart = i
+			} else {
+				sectionStart = -1
+			}
+			continue
+		}
+		if sectionStart < 0 {
+			continue
+		}
+		if strings.HasPrefix(t, "team_id") && strings.Contains(t, teamID) {
+			return sectionStart
+		}
+	}
+	return -1
+}
+
 // saveGlobalTheme rewrites the [appearance] theme line in config.toml.
 // If the file has no theme line, it appends a new [appearance] section.
 // Existing comments and ordering are preserved (textual rewrite, not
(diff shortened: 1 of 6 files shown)


gammons:
The bug you found is real and it's a P1 — I reproduced it on clean main:

```
[workspaces.T01ABCDEF]        <- appended by the save helper
version_ts = "111"
[workspaces.work]
team_id = "T01ABCDEF"

Load: workspaces "T01ABCDEF" and "work" both reference team_id "T01ABCDEF"
```

`saveWorkspaceVersionTS`, `saveWorkspaceTheme` and `saveWorkspaceWidth` all match only `[workspaces.<tomlKey>]` exactly and fall through to appending `[workspaces.<teamID>]`, and `workspaceTOMLKey` returns the raw team ID whenever the in-memory config has no block for that team — which happens any time the file changed after load. Then `resolveWorkspaceKeys` hard-errors and slk won't start. Good find, and deduplicating the three copy-pasted scanners into `findWorkspaceSectionStart` is the right cleanup.

**But the fix introduces two silent-corruption paths that are worse than the loud error it replaces.** I reproduced both on your branch.

**1. Saved themes are silently discarded.** `findWorkspaceSectionStart` (`save_theme.go:26-33`) loops lines outer, headers inner, so it returns whichever header appears **first in the file** — not the priority order its own doc comment at `save_theme.go:12-13` promises. When the leftover block precedes the slug block, every save lands in the leftover, and `mergeWorkspacePrefs` (`workspaces.go:283-288`) then gives the slug's non-empty value precedence:

```
before: [workspaces.T01ABCDEF] version_ts="111" ; [workspaces.work] team_id=..., theme="dracula"
        saveWorkspaceTheme(path, "work", "T01ABCDEF", ..., "nord")
after:  [workspaces.T01ABCDEF] theme = "nord"     <- wrong block
        [workspaces.work]      theme = "dracula"
Load -> effective theme = "dracula"               <- the user's theme change vanished
```

On main this same call correctly writes into `[workspaces.work]`. That's a regression. The doc comment is right; the loop order is wrong — invert it.

**2. Cross-workspace config corruption.** `save_theme.go:51` uses `strings.Contains(t, teamID)`. `team_id = "T01ABCDEFGH"` *contains* `T01ABCDEF`, so a save for one team writes into a **different workspace's** block:

```
before: [workspaces.other] team_id = "T01ABCDEFGH"  theme = "dracula"
        saveWorkspaceVersionTS(path, "T01ABCDEF", "T01ABCDEF", ..., "999")
after:  [workspaces.other]
        version_ts = "999"       <- wrong workspace's block mutated
        team_id = "T01ABCDEFGH"
```

Needs an exact parse of the `team_id` value — strip `=`, quotes, and any trailing `#` comment — not a substring match.

**3. Silent field loss on merge.** `mergeWorkspacePrefs` (`workspaces.go:282-301`) covers all 7 fields of `config.Workspace` today, so nothing is dropped right now. But when the leftover block sets a field the slug also sets, the leftover value is discarded with no log line where main used to produce a loud error. Add a `debuglog` note. And the manual field enumeration has no compile-time protection — the next field added to `Workspace` gets silently dropped. A reflection-based test asserting every field is handled would prevent that.

**4. It never self-heals.** I confirmed `Load` doesn't rewrite the file, so this is idempotent across boots — good. But it also means the leftover block persists on disk forever, and combined with (1) it keeps stealing writes. Consider having `Load` (or a one-shot migration) rewrite the file once via the existing `writeConfigAtomic` to remove the merged block.

Your existing tests are meaningful and `TestConfiguredTeamIDs_WorksOnAConfigThatFailsValidation` pins a design decision worth pinning. Neither covers the two orders that actually break, though — leftover-block-first, and substring-colliding team IDs. Both repros above are trivial to add.

Smaller: `resolveWorkspaceKeys` mutates `prepared` while ranging over it (`workspaces.go:261-273`). Legal here since only existing keys are updated and slug keys are `continue`d, but it deserves a comment. And `findWorkspaceSectionStart` living in `save_theme.go` while three files use it is a bit arbitrary — `save_workspace_section.go` would be clearer.

Longer term I think the right fix is narrower: make the save helpers resolve the target section by parsing `team_id` correctly so the duplicate is never created, and treat the `Load`-side merge as a one-time repair for already-broken files (and say so in the comment).

Your CI failure was **not your fault** — that's a flaky test on main in `internal/slack/membership`, unrelated to your diff. Fixed in #171; rebase after that lands.

gammons:
#171 has landed and I've re-run CI here — **`test` is green now.** That failure was a flaky test on main in `internal/slack/membership`, unrelated to your diff.

Note that #171 also enabled `gofmt` as an enforced lint check, so run `gofmt -w` over your changes before pushing.

My review above still stands, and the two silent-corruption paths are the blocking part: the loop order in `findWorkspaceSectionStart` (saves landing in the wrong block and being discarded on load) and the `strings.Contains` team-ID match (writing into a different workspace's block). Both have reproductions in that comment.

PR #168, opened 2026-08-31
feat(sidebar): hoist Direct Messages above other sections

agustif:
## Why

Direct Messages sat at the bottom of the rail (after custom sections in config mode, and at the end of Slack's linked-list order). They are the daily-driver conversations and belong under Threads.

## What

- Config mode: **DMs → custom sections → Apps → Channels**.
- Slack mode: keep the provider linked list, but hoist `direct_messages` to the top.

Threads remains the synthetic top row in this PR (Activity is the next PR).

## Test plan

- [x] `go test ./internal/ui/sidebar/`
- [ ] Confirm the sidebar lists Direct Messages immediately under Threads, then custom sections, then Apps/Channels.

**Merge after #167** (stack 3/4). Contains PR1 + PR2 + this commit.

### DIFF
--- a/cmd/slk/bootstrap_adapters.go
+++ b/cmd/slk/bootstrap_adapters.go
@@ -455,6 +455,39 @@ func bootConversations(res *bootstrap.Result) []slack.Channel {
 	return out
 }
 
+// fillIMUsers copies counterparty user IDs from userBoot's ims[] onto
+// IM conversations that lack them.
+//
+// users.conversations is the sidebar's primary source, but its IM
+// objects are not guaranteed to carry `user` — Slack's conversation
+// payload sometimes omits it, and slack-go then leaves Channel.User
+// empty. buildChannelItem names a DM from that field (falling back to
+// the raw user ID), and an empty User means the sidebar row is blank
+// and the unresolved-DM sweep has nothing to fetch. userBoot's ims[]
+// always include `user`; overlaying them is how a DM gets a name when
+// GetChannels succeeded. Existing User values are left alone.
+func fillIMUsers(channels []slack.Channel, ims []boot.IM) {
+	if len(channels) == 0 || len(ims) == 0 {
+		return
+	}
+	byID := make(map[string]string, len(ims))
+	for _, im := range ims {
+		if im.UserID != "" {
+			byID[im.ID] = im.UserID
+		}
+	}
+	for i := range channels {
+		uid, ok := byID[channels[i].ID]
+		if !ok {
+			continue
+		}
+		channels[i].IsIM = true
+		if channels[i].User == "" {
+			channels[i].User = uid
+		}
+	}
+}
+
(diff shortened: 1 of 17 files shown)


gammons:
The code here is clean — 46 changed lines in one file, the rest is assertion churn, no smuggled refactors. Credit for that; it's a genuinely reviewable diff.

My problem is with the product decision, not the implementation.

**This overrides the user's Slack-side section order.** `internal/ui/sidebar/model.go:459-472` splits the provider's linked list and force-hoists `direct_messages`. But `wiki/Configuration.md:115-118` states that slk "reads the user's actual sidebar sections — names, emoji, **linked-list order**, and channel membership — directly from Slack and keeps them live," and `docs/STATUS.md:104` says the same. Slack's DM section position *is* user-configurable in the official client. So after this, slk shows a different order than the Slack app the user just came from — which is the exact failure mode this subsystem exists to prevent. The PR doesn't update `wiki/` or `docs/` to reflect the change.

**And it removes control in config mode too.** Moving `defaultDMSection` above the `customs` loop (`model.go:147-155`) means **no value of `order` can place a custom section above DMs** anymore. `SectionOrder` is documented at `wiki/Configuration.md:74-76` ("Lower numbers sort higher"), and `"Direct Messages"` is a synthesized fallback name, not addressable from `[sections.*]`. Someone who deliberately set `[sections.Alerts] order = 1` to keep alerts at the top loses that permanently with no way to get it back.

The tests can't help here by construction — all three files just flip existing assertions (`TestOrderedSections_SlackMode_HonorsLinkedListOrder` becomes `..._HoistsDirectMessages`, and so on). The previous tests encoded the previous intent just as firmly. That's expected for a pure reordering, which is exactly why this needed a design conversation rather than a diff.

**DMs-first is a defensible default and I'm not opposed to it.** But it needs to be a setting, not a hardcode:

1. Open an issue proposing the change so we can agree on the shape (per `README.md:107-110` — this is a global, non-configurable change to the primary navigation surface, which qualifies as a large feature addition).
2. Make it configurable — something like `[general] dm_section_position = "top" | "slack" | "bottom"` — defaulting to `"slack"` in Slack-native mode so the mirroring contract holds by default.
3. In config mode, express DMs as an orderable pseudo-section so `SectionOrder` can still place a custom section above it.
4. Update `wiki/Configuration.md` and `docs/STATUS.md:104` for whatever the final default is.
5. Add a Slack-mode test where the provider already returns `direct_messages` first — the hoist should be a no-op and the rest of the order preserved. Also worth covering multiple `direct_messages`-typed sections.

Leaving this open for now. If you'd rather start from the issue, close it and I'll pick up the discussion there.

One note on sequencing: if #167's blocking name-fetch gets dropped (which I'm asking for over there) and this lands, unresolved DMs would render as raw user IDs *at the top* of the sidebar. So #167's `fillIMUsers` fix should land first regardless of what happens here.

PR #169, opened 2026-08-31
feat(activity): Slack Activity inbox

agustif:
## Why

Slack's Activity inbox (mentions, DMs, reactions, …) is the other half of "what needs my attention" next to Threads. slk had no view for it.

## What

- New `activity.feed` / `activity.views` client (unofficial Slack protocol, same class as the rest of slk — it can change without notice).
- Synthetic **Activity** row above Threads with an `activity_v2` unread badge from `client.counts`.
- Message pane: filter / sort / unread-only, compact and detailed density, jump-to-message.
- Prefs in `[activity]` in config.toml.
- Docs: README, wiki Features / Configuration / Keybindings.

## Test plan

- [x] `go test ./internal/config/ ./internal/slack/ ./internal/bootstrap/ ./internal/ui/ ./internal/ui/sidebar/ ./internal/ui/activityview/ ./cmd/slk/`
- [ ] Open Activity from the sidebar; feed loads; unread badge matches Slack.
- [ ] Cycle filters, toggle unread-only, jump to a mention/DM.
- [ ] Reconnect: badge refreshes from `client.counts`.

**Merge after #168** (stack 4/4). Contains the full stack. Testdata is sanitized API JSON (no tokens).

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@
 - Real-time messages, edits, deletes, reactions, typing indicators
 - Inline images (kitty graphics / sixel / half-block fallback) with full-screen preview
 - Threads side panel + a workspace-wide threads view
+- Activity inbox (Slack's Activity tab: recents, filters, unread badge)
 - Smart paste: clipboard images, file paths, or text — multiple attachments + caption in one send
 - Slack-native sidebar sections, kept live; or glob-based config sections
 - Automatic auth from the Slack desktop app — no tokens to copy, no Slack App required
--- a/cmd/slk/bootstrap_adapters.go
+++ b/cmd/slk/bootstrap_adapters.go
@@ -455,6 +455,39 @@ func bootConversations(res *bootstrap.Result) []slack.Channel {
 	return out
 }
 
+// fillIMUsers copies counterparty user IDs from userBoot's ims[] onto
+// IM conversations that lack them.
+//
+// users.conversations is the sidebar's primary source, but its IM
+// objects are not guaranteed to carry `user` — Slack's conversation
+// payload sometimes omits it, and slack-go then leaves Channel.User
+// empty. buildChannelItem names a DM from that field (falling back to
+// the raw user ID), and an empty User means the sidebar row is blank
+// and the unresolved-DM sweep has nothing to fetch. userBoot's ims[]
+// always include `user`; overlaying them is how a DM gets a name when
+// GetChannels succeeded. Existing User values are left alone.
+func fillIMUsers(channels []slack.Channel, ims []boot.IM) {
+	if len(channels) == 0 || len(ims) == 0 {
+		return
+	}
+	byID := make(map[string]string, len(ims))
+	for _, im := range ims {
+		if im.UserID != "" {
+			byID[im.ID] = im.UserID
+		}
+	}
+	for i := range channels {
+		uid, ok := byID[channels[i].ID]
+		if !ok {
+			continue
+		}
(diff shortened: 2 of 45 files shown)


gammons:
@agustif — thank you for this, it's a lot of careful work and several parts of it are genuinely better than what I'm merging. I'm going with #109 though, and I want to be straight with you about why rather than leave you guessing.

**1. Row content.** #109 hydrates message bodies via `messages.list` and renders the actual text. This renders metadata only — `previewText` (`internal/ui/activityview/model.go:815-847`) returns `"direct message"`, `"reply from Alice"`, `":tada:"` — and there's no `messages.list` call anywhere in the PR. That's specifically what I asked #109 to change back in July, with screenshots. Related: `flattenActivityEntry` doesn't set `ActorID` for `dm` or `thread_v2`, so those rows have no sender name either.

**2. All bot DMs are silently dropped.** `flattenActivityEntry` (`internal/slack/activity.go:350-392`) has no `bot_dm_bundle` case. The `default` arm reads `inner.Message.Channel`, which is empty for that shape (the real payload puts it at `bundle_info.payload.message`), and the `ChannelID == ""` guard then discards the item at line 388-390. I ran your parser against #109's capture of a real `bot_dm_bundle` entry and it returns **0 items**. GitHub, PagerDuty, Jira, Slackbot — a large share of a real Activity tab — just don't appear. Your 91-line fixture doesn't include that type, which is why the tests are green.

**3. Process.** #109 was open a month, went through two rounds of my review, and I'd already told the author it was ready to merge. I'm not going to leapfrog that.

Also, for the record: `channelGlyph` (`activityview/model.go:935`) is dead code and that's your lint failure. Your test failure was *not* your fault — that's a flaky test on main which I've fixed in #171.

**Three things in here I want, and will happily take as standalone PRs against main, credited to you:**

1. **The `activity_v2` badge from `client.counts`** — the `GetCounts`/`CountsSnapshot` refactor in `internal/slack/client.go`. This is strictly better than deriving the badge from the fetched feed: zero extra requests, and it's correct at cold start and after reconnect. #109's badge reads 0 until you open the view. **I'd take this one today.**
2. **`activity.views` support** — custom tabs (Unreads / Reactions / VIP) with the `f`/`F`/`s` cycling and the `[activity]` config block. Well tested, and the tests asserting exact request form shape (`TestBuildActivityFeedForm_*`) are the right way to pin an undocumented endpoint.
3. **Generation-guarded fetches** — `ActivityFeedLoadedMsg.Gen`. #109 needs this and doesn't have it. Your `TestApp_ActivityFeedLoadedDropsStaleGen` is a good test.

Two pieces of process feedback, offered constructively:

**Please stop stacking feature PRs.** This is 45 files because it carries #166, #167 and #168 underneath it. Even measured against #168 it's still 38 files / +3220. Reviewing it means reviewing four PRs at once, and that's a large part of why it sat. I'm reviewing those three separately on their own merits.

**And per CONTRIBUTING, open an issue first for something this size.** #12 was the right place — and doing that would have surfaced #109 before you built this, which would have saved you a lot of work. That's the actual purpose of the rule, not bureaucracy.

If you want to fix `VIP: len(e.Priority) > 2` (`activity.go:359`) while you're in there — inferring VIP from the byte length of raw JSON means an explicit `"priority": null` gives `len == 4` and renders a spurious `vip` chip on every row. A typed decode would be better.

Closing as superseded. I'd genuinely like those three follow-ups.

PR #176, opened 2026-09-04
fix(mrkdwn): treat a single asterisk pair as bold, matching Slack's own mrkdwn

rfist:
## Problem

Typing `*bold*` in the composer sent literal, unstyled asterisks to Slack.

slk renders `*bold*` as bold in its own compose box and message view, because that is what Slack's mrkdwn means — a single asterisk pair is Slack's native bold syntax. But the outgoing converter treated the same input as *literal text*: `walkAsteriskLiteral` emitted `*`, the inner run, and `*` as plain text elements with no style of their own. So a message that looked bold while you were writing it arrived in real Slack with the asterisks showing.

The cause is a CommonMark/mrkdwn mismatch. Goldmark parses `*x*` as `ast.Emphasis` — CommonMark *italic* — and the walker only distinguished the delimiter to avoid emitting `_x_` for it:

```go
if w.emphasisDelimiter(n) == '_' {
    w.walkItalic(n)
} else {
    w.walkAsteriskLiteral(n)   // preserved the asterisks as text
}
```

## Fix

One line: the asterisk branch calls `walkBold` instead, so a single asterisk pair produces the same `rich_text` styling as `**x**`.

```go
if w.emphasisDelimiter(n) == '_' {
    w.walkItalic(n)
} else {
    w.walkBold(n)
}
```

`walkAsteriskLiteral` is deleted along with it — nothing else called it.

Scope is deliberately narrow. Only single-asterisk emphasis changes: `_italic_` still emits italic, `**bold**` and `__bold__` are untouched, and the mrkdwn string output is unchanged in every case (`*hello*` still round-trips to `*hello*`). Only the block side gains the style it should always have had.

## About the test diff

**This removes 74 test lines, and none of it is lost coverage.** Three tests asserted the *old* behaviour on purpose and had to be inverted rather than deleted:

| Before | After |
|---|---|
| `TestConvert_AsteriskPreservedAsLiteral` | `TestConvert_SingleAsteriskIsBold` |
| `TestConvert_NestedItalicAroundAsteriskLiteral` | `TestConvert_NestedItalicAroundBold` |
| `TestConvert_NestedAsteriskLiteralAroundItalic` | `TestConvert_NestedBoldAroundItalic` |

Each got shorter because the assertion got simpler. The old ones had to walk a variable number of text elements, concatenate them, and check that the asterisks carried no style of their own while inheriting any enclosing one. The new ones assert a single element with the expected style — `*hello*` is one bold element, `_*both*_` and `*_both_*` are each one element carrying italic and bold.

Worth noting that the old test's own comment argued for this change:

> `*x*` in CommonMark is italic, but in Slack mrkdwn it's bold. We preserve the asterisks as literal characters so users who type Slack-style `*bold*` don't get it converted to `_italic_`.

The goal was already right — stop `*bold*` becoming `_italic_`. Making the asterisks literal stopped the wrong conversion without doing the right one. This finishes the job.

Untouched and still passing: `TestConvert_BoldDoubleAsterisk`, `TestConvert_BoldDoubleUnderscore`, `TestConvert_ItalicUnderscore`, `TestConvert_ItalicContainingBold`, `TestConvert_AsteriskRoundTripStable`, and the link/mention/strikethrough style-inheritance tests.

## Documentation

`wiki/Features.md` documented the old behaviour explicitly — "Single-asterisk emphasis (`*x*`) is preserved as literal text since it conflicts with Slack mrkdwn bold" — so it is corrected in the same commit. The preceding sentence already said that already-mrkdwn syntax passes through unchanged; it now also says that it receives the corresponding rich-text styling, which I verified holds for `*bold*`, `_italic_` and `~strike~` alike.

The `docs/superpowers/` references to literal asterisks are left alone: those are the dated design plan and spec from the original CommonMark-compose work, i.e. a record of what was decided then, not current user documentation.

## Test plan

- Confirmed the tests fail against the old code: reverting `walk.go` to main fails exactly `TestConvert_SingleAsteriskIsBold`, `TestConvert_NestedItalicAroundBold` and `TestConvert_NestedBoldAroundItalic`, and nothing else.
- `go test ./...`, `go vet ./...`, `gofmt -l .` clean, `go build ./...`.
- Rebased onto current main.

### DIFF
--- a/internal/slack/mrkdwn/convert_test.go
+++ b/internal/slack/mrkdwn/convert_test.go
@@ -186,29 +186,25 @@ func TestConvert_ItalicUnderscore(t *testing.T) {
 	}
 }
 
-func TestConvert_AsteriskPreservedAsLiteral(t *testing.T) {
-	// *x* in CommonMark is italic, but in Slack mrkdwn it's bold.
-	// We preserve the asterisks as literal characters so users who
-	// type Slack-style *bold* don't get it converted to _italic_.
+func TestConvert_SingleAsteriskIsBold(t *testing.T) {
+	// *x* in CommonMark is italic, but Slack's own mrkdwn treats a
+	// single asterisk pair as bold. slk treats it as bold (same as
+	// **x**) so a message that renders bold in slk's own compose/
+	// message view also renders bold in real Slack.
 	mr, blk := Convert("*hello*")
 	if mr != "*hello*" {
 		t.Errorf("mrkdwn = %q, want %q", mr, "*hello*")
 	}
-	// Block side: text elements concatenate to "*hello*", no italic.
 	sec := blk.Elements[0].(*slack.RichTextSection)
-	var got string
-	for _, el := range sec.Elements {
-		te, ok := el.(*slack.RichTextSectionTextElement)
-		if !ok {
-			t.Fatalf("element is %T, want only text elements", el)
-		}
-		if te.Style != nil && (te.Style.Italic || te.Style.Bold) {
-			t.Errorf("element %+v has style %+v, want no italic/bold for literal asterisks", te, te.Style)
-		}
-		got += te.Text
+	if len(sec.Elements) != 1 {
+		t.Fatalf("elements = %d, want 1: %s", len(sec.Elements), blockJSON(blk))
 	}
-	if got != "*hello*" {
-		t.Errorf("concatenated text = %q, want %q", got, "*hello*")
+	te := sec.Elements[0].(*slack.RichTextSectionTextElement)
+	if te.Text != "hello" {
+		t.Errorf("text = %q, want %q", te.Text, "hello")
+	}
+	if te.Style == nil || !te.Style.Bold {
+		t.Errorf("expected Style.Bold = true, got %+v", te.Style)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


gammons:
Reviewed this as a behavior change, not just a fix, because it inverts a deliberate-looking decision — the old code had a comment explaining it preserved `*x*` as literal text "so users who type Slack-style *bold* don't get it converted to _italic_."

Your counter-argument is the right one, and I checked it: Slack's own mrkdwn treats a single asterisk pair as **bold**, and slk's message view already renders `*word*` as bold. So the old behavior produced the worst of three outcomes — the user typed `*bold*`, watched it render bold while composing, and then the message arrived in Slack with literal asterisks showing. Preserving-literal avoided a wrong italic conversion but introduced a wrong literal one. Making it bold is consistent with both surfaces.

Verified the scope is as narrow as you claim:

- `**bold**` parses as `ast.Strong`, a different node type, so the `walkBold` change can't touch it. `_italic_` is unchanged. The mrkdwn string output is byte-identical (`*hello*` round-trips to `*hello*`), so nothing on the wire-format side moved.
- The nested cases (`_*both*_`, `*_both_*`) now correctly produce italic+bold on "both", and the rewritten tests assert that.
- Deleting `walkAsteriskLiteral` is safe — nothing else called it.

The test inversions are handled honestly (renamed, not deleted; the three flipped tests are listed in your table with before/after). That's the discipline that makes a -74-line test diff reviewable.

Full suite green: build, vet, gofmt clean, 54 packages under `-race`.

Merging.

Issue #165, opened 2026-08-29
Emoji pickers refold every entry per keystroke, causing 70-164 ms input stalls at 64k entries


YousefHadder:
## What happens

With the full workspace emoji list loaded, rare or no-match queries visibly stall input in both emoji pickers. At roughly 64,000 searchable entries, independently measured full scans took approximately:

- 69-88 ms per keystroke in the compose emoji picker.
- 116-164 ms per keystroke in the reaction picker.

This is latent on [`c782404`](https://github.com/gammons/slk/commit/c782404508d40e6641d1dc7cff08a5fbd2f28c16). `conversations.view` supplies only the custom emoji used in the boot history, and the bootstrap bug addressed by [PR #160](https://github.com/gammons/slk/pull/160) currently prevents the later workspace-wide [`emoji.list`](https://api.slack.com/methods/emoji.list) fetch. Fetching the complete list makes this cost user-visible, so the filtering path should be optimized alongside that fix.

## Measurements

The benchmark used approximately 62,000 synthetic custom emoji, producing roughly 64,000 total entries with the bundled catalog. No names, URLs, messages, or other content from the private workspace were used. A rare six-character query was entered one character at a time to exercise the full-scan path. Results below are medians of three benchmark runs, with ten iterations per run, on Darwin/arm64.

| Query | Compose current | Compose precomputed | Reaction current | Reaction precomputed |
|---|---:|---:|---:|---:|
| 1 character | 80.8 ms | 0.146 ms | 160.6 ms | 0.375 ms |
| 2 characters | 82.9 ms | 0.173 ms | 155.5 ms | 0.615 ms |
| 3 characters | 80.2 ms | 0.172 ms | 153.1 ms | 0.575 ms |
| 4 characters | 69.0 ms | 0.157 ms | 116.9 ms | 0.557 ms |
| 5 characters | 72.5 ms | 0.161 ms | 116.5 ms | 0.525 ms |
| 6 characters | 80.7 ms | 0.157 ms | 117.7 ms | 0.506 ms |
| **Typing total** | **466 ms** | **0.97 ms** | **820 ms** | **3.15 ms** |

The current compose scan allocated approximately 563 MB across 386,000 allocations per query. The reaction scan allocated approximately 1.13 GB across 773,000 allocations per query. Those totals are dominated by the per-entry transformation chains.

The precomputed comparison loops allocated zero bytes. A production implementation would still fold the user's query once per keystroke, which is negligible compared with folding every candidate.

[...]

PR #177, opened 2026-09-04
perf(reactionpicker): fold candidate names once

YousefHadder:
## Summary

- fold each reaction-picker candidate name once and reuse it for prefix and substring checks
- cover substring selection through the exported `SetCustomEmoji`, `Open`, and `HandleKey` seam, observing the selected `ReactionResult`
- benchmark a full-scan query over 10,000 custom emoji through the same input seam

This is the first of two independent performance follow-ups requested after PR #174 / issue #165.

## Benchmark

Apple M3 Max, darwin/arm64:

```console
go test ./internal/ui/reactionpicker -run '^$' -bench '^BenchmarkFilterByQuery$' -benchmem -count=6 -benchtime=500ms
```

Observed median: **604,271 -> 341,907 ns/op** (**43.4% lower**). Memory remained **0 B/op, 0 allocs/op** before and after.

## Testing

- `go test ./internal/ui/reactionpicker -run 'TestSubstringSearchSelectsCustomEmoji|TestFilterByQuery'`
- `go test -race ./internal/ui/reactionpicker`
- benchmark command above

Refs #165

### DIFF
--- a/internal/ui/reactionpicker/bench_test.go
+++ b/internal/ui/reactionpicker/bench_test.go
@@ -1,0 +1,24 @@
+package reactionpicker
+
+import (
+	"fmt"
+	"testing"
+)
+
+func BenchmarkFilterByQuery(b *testing.B) {
+	customs := make(map[string]string, 10_000)
+	for i := 0; i < 10_000; i++ {
+		name := fmt.Sprintf("candidate_%05d", i)
+		customs[name] = "https://emoji.example.com/" + name + ".gif"
+	}
+
+	m := New()
+	m.SetCustomEmoji(customs)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		m.Open("C123", "1234.5678", nil)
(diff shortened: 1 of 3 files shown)


gammons:
Clean. This collapses the double `Fold` per prefix miss that I flagged in #165, and the benchmark is the right call — it's now measuring the real path, so the next person can't accidentally reintroduce a per-candidate allocation without it showing up.

`TestSubstringSearchSelectsCustomEmoji` is a good end-to-end check: it drives the picker through actual keystrokes and asserts the substring result is selectable, which exercises more than a unit test of `filter` would.

One note, not a change: since #174 landed, this loop was already cheap (ASCII names now fold to `strings.ToLower`, zero-alloc), so the double-fold is more about clarity than measured latency here. The win is real but the `Fold` fast path is doing most of the heavy lifting — worth keeping that in the description's context so the benchmark numbers aren't read as if the pre-#174 cost still applied.

Verified: build, vet, gofmt clean, 54 packages under `-race`.

Merging.

PR #178, opened 2026-09-04
perf(messages): fold ASCII highlight segments once

YousefHadder:
## Summary

- Replace the private hand-rolled ASCII lowering path with segment-level `internal/text.Fold` now that PR #174 centralizes the ASCII fast path.
- Fold valid ASCII segments once, then split the result back into the existing one-entry-per-rune mapping.
- Keep the per-rune `text.Fold` fallback for non-ASCII and invalid UTF-8 so highlight indices and behavior remain unchanged.

## Behavior coverage

Behavior-level tests exercise the `HighlightSearchTerms` package seam, including decomposed combining-mark mapping (`Cafe\u0301`) to preserve the original rune-to-highlight relationship.

## Observed benchmark

Environment: Apple M3 Max, darwin/arm64.

Protocol:

```sh
go test ./internal/ui/messages -run '^$' -bench '^BenchmarkHighlightSearchTerms_ASCII$' -benchmem -count=6 -benchtime=500ms
```

| Implementation | Median ns/op | Observed range ns/op | B/op | allocs/op |
| --- | ---: | ---: | ---: | ---: |
| Original hand-rolled baseline | 2,789 | 2,755-3,318 | 5,008 | 185 |
| Final coarse `text.Fold` path | 1,855 | 1,837-1,921 | 4,304 | 9 |

In these measurements, the median was 33.5% lower and allocations decreased from 185 to 9 per operation. These are observed results on the environment above, not universal performance claims.

## Context

This is the independent message-highlighting follow-up requested after PR #174 / issue #165. It is separate from the reaction-picker follow-up in PR #177.

Refs #165

### DIFF
--- a/internal/ui/messages/highlight.go
+++ b/internal/ui/messages/highlight.go
@@ -129,16 +129,17 @@ func HighlightSearchTerms(s string, terms []string, hlStart, hlEnd string) strin
 		}
 		runes := []rune(sg.text)
 		folded := make([]string, len(runes))
-		for i, r := range runes {
-			if r < utf8.RuneSelf {
-				// ASCII fast path: text.Fold allocates a transform
-				// chain per call (see fold.go); for ASCII, folding is
-				// just lowercasing.
-				if r >= 'A' && r <= 'Z' {
-					r += 'a' - 'A'
-				}
-				folded[i] = string(r)
-			} else {
+		if len(runes) == len(sg.text) && utf8.ValidString(sg.text) {
+			// ASCII is one byte per rune, so folding the segment once keeps
+			// each byte aligned with the corresponding original rune.
+			foldedText := text.Fold(sg.text)
+			for i := range folded {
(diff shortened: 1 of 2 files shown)


willinghamlamont00-a11y:
Great

gammons:
This is exactly right, and it's the right way to retire that workaround.

The guard `len(runes) == len(sg.text)` is the load-bearing detail — it means the segment is one byte per rune, i.e. pure ASCII, which is precisely the case where `Fold` is now byte-preserving (lowercasing only), so `foldedText[i]` stays aligned with the original rune. Anything else falls back to per-rune folding, and the decomposed-accent test (`Cafe\u0301`) pins the case where folding would change the byte count. That's the regression I'd have worried about silently breaking, and you've tested it directly.

This is the follow-up I was hoping you'd take. Verified: build, vet, gofmt clean, full `-race` suite green (54 packages).

Merging.

PR #196, opened 2026-09-08
fix (slackdesktop): detect MSIX/Store Slack installations on Windows

### LINKED ISSUE #183: MSIX/Microsoft Store Slack Installations Not Detected
## Problem

The storage location of Slack can differ based on the installation method used.

1. Legacy installer (SlackSetup.exe): `%APPDATA%\Slack`
2. MSIX or Microsoft Store: `%LOCALAPPDATA%\Packages\com.tinyspeck.slackdesktop_8yrtsj140pw4g\LocalCache\Roaming\Slack`

Currently, slk attempts to read the signed-in workspaces only from the `%APPDATA%\Slack` folder, meaning MSIX and Store installations of the software won't be detected.

## Expected Behaviour

Slack desktop app in Windows should be automatically detected, no matter the method of installation.

## Steps to Reproduce

1. Uninstall Slack, if you already have it.
2. Delete the `%APPDATA%\Slack` folder, to ensure Slack doesn't read and use it during installation.
3. Install Slack using the MSIX installer or from Microsoft Store.
4. Run `slk --add-workspace`.

## Relevant Code

- `internal/slackdesktop/configdir.go`: `configDirForOS`

Allison-E:
Fixes #183.

## The Bug

Slack installed using the MSIX installer or Microsoft Store is not detected by slk. The only path checked is `%APPDATA%/Slack`, which corresponds to the Slack instance installed via the Legacy Installer (SlackSetup.exe).

## Approach

Try to detect Slack's folder, checking `%APPDATA%/Slack` first, and then `%LOCALAPPDATA%\Packages\com.tinyspeck.slackdesktop_8yrtsj140pw4g\LocalCache\Roaming\Slack` if the first doesn't exist.

## Bug Fixed Along the Way

When a Slack process is running on Windows, the Cookie database is exclusively locked. The attempt to open this file in `copyToTemp` returns the OS-level error `windows.ERROR_SHARING_VIOLATION (32)`, however, the error message returned to the user is generic: "slack cookie database not found".

In this update, the error message returned to the user is improved by first detecting this Cookie database lock by the running Slack process and informing the user to close Slack before trying again.

## Testing
- Wrote tests for detecting a lock on the Cookie database on Windows and for ensuring the MSIX/Store-installed Slack instances are detected
- `make test`, `go vet ./...`, `make build` on Windows and Linux

## Side Note
- Detection of exclusive locks on the Cookie database is detected on Windows only. Implementations for other OS' are not done.

### DIFF
--- a/cmd/slk/onboarding.go
+++ b/cmd/slk/onboarding.go
@@ -141,6 +141,8 @@ func desktopErrorMessage(err error) string {
 		return "Could not decrypt the Slack session cookie: " + err.Error() +
 			". If you have had more than one Slack build installed (App Store and standalone), " +
 			"sign out of the one you no longer use. Otherwise please file an issue with your OS + Slack version."
+	case errors.Is(err, slackdesktop.ErrCookieLocked):
+		return "Your Slack cookie is locked by a running process of Slack. Close Slack and try again."
 	default:
 		return "Could not read Slack desktop session: " + err.Error()
 	}
--- a/internal/slackdesktop/configdir.go
+++ b/internal/slackdesktop/configdir.go
@@ -11,12 +11,17 @@ import (
 func configDirForOS(goos string, getenv func(string) string, exists func(string) bool) string {
 	switch goos {
 	case "windows":
-		first := filepath.Join(getenv("APPDATA"), "Slack")
-		second := filepath.Join(getenv("LOCALAPPDATA"), "Packages", "com.tinyspeck.slackdesktop_8yrtsj140pw4g", "LocalCache", "Roaming", "Slack")
-		if exists(first) {
-			return first
+		var candidates = [...]string {
+			filepath.Join(getenv("APPDATA"), "Slack"),
+			filepath.Join(getenv("LOCALAPPDATA"), "Packages", "com.tinyspeck.slackdesktop_8yrtsj140pw4g", "LocalCache", "Roaming", "Slack"),
 		}
-		return second
+		for _, c := range candidates {
+			if exists(c) {
+				return c
+			}
+		}
+		// Fall back to the first candidate if none was found.
+		return candidates[0]
 	case "darwin":
 		home := getenv("HOME")
 		first := filepath.Join(home, "Library", "Application Support", "Slack")
--- a/internal/slackdesktop/configdir.go
+++ b/internal/slackdesktop/configdir.go
@@ -11,10 +11,9 @@ import (
 func configDirForOS(goos string, getenv func(string) string, exists func(string) bool) string {
 	switch goos {
 	case "windows":
-		var candidates = [...]string {
-			filepath.Join(getenv("APPDATA"), "Slack"),
-			filepath.Join(getenv("LOCALAPPDATA"), "Packages", "com.tinyspeck.slackdesktop_8yrtsj140pw4g", "LocalCache", "Roaming", "Slack"),
-		}
+		var candidates []string
+		candidates = append(candidates, filepath.Join(getenv("APPDATA"), "Slack"))
+		candidates = append(candidates, filepath.Join(getenv("LOCALAPPDATA"), "Packages", "com.tinyspeck.slackdesktop_8yrtsj140pw4g", "LocalCache", "Roaming", "Slack"))
 		for _, c := range candidates {
 			if exists(c) {
 				return c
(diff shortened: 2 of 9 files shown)


gammons:
@Allison-E Welcome, and thanks for this — it's a real gap, well found, and the MSIX path matches Microsoft's documented location for Store-installed desktop apps. The detection logic is correct and consistent with the candidate-list approach we just landed for flatpak/snap in #128. The two `configdir` tests cover both branches properly.

**But the lock-handling half is currently inert — and it's a one-word fix.**

The `IsFileLockError` implementation is right: build-tagged, checks `windows.ERROR_SHARING_VIOLATION` through `errors.AsType`, and it's the most-tested part of the change (141 lines). But look at where its result goes in `internal/slackdesktop/cookiedb.go:56-63`:

```go
if IsFileLockError(err) {
	return "", ErrCookieDBMissing   // <-- this is the bug
} else {
	return "", ErrCookieDBMissing
}
```

Both branches return the same error. So the lock *is* detected correctly, and then discarded.

The consequence is that the message you added at `cmd/slk/onboarding.go:144` — "Your Slack cookie is locked by a running process of Slack. Close Slack and try again." — is **unreachable**. You defined `ErrCookieLocked` and wired it into onboarding, but nothing ever returns it. A user with a locked cookie DB still gets "cookie database not found" (from `onboarding.go:129`), which points them at the wrong fix entirely.

I traced the propagation to make sure the fix is as small as it looks: `copyToTemp`'s error passes through `readCookieRow` (`cookiedb.go:27`) unwrapped, then `cookie.go:114`, then up to onboarding, and `errors.Is(err, ErrCookieLocked)` matches there. I verified the two errors are correctly distinguished. So literally:

```go
if IsFileLockError(err) {
	return "", ErrCookieLocked   // not ErrCookieDBMissing
}
```

**Why it shipped this way:** your lock tests exercise `IsFileLockError` in isolation (correctly — you can't portably trigger a real OS lock in a unit test), but nothing asserts that a locked file *surfaces* as `ErrCookieLocked` at the `copyToTemp` boundary. That's the gap that let the discarded result through. Worth adding a test there: stub the open to return a lock-shaped error and assert `copyToTemp` returns `ErrCookieLocked`, which would have caught this.

Two smaller things:

- The `else` after `return` is redundant — `gofmt`/staticcheck won't flag it, but a plain second `return` reads better.
- The `configdir.go` change is fine, but we just landed #128 which refactored this exact function to a candidate list. If you rebase, `case "windows":` becomes another candidate line, which fits your two-branch logic naturally — worth aligning to that shape.

This is close, and the hard part (getting the Windows locking semantics right) is done correctly. Fix the error mapping and the test, and I'll take it.

Allison-E:
@gammons Thank you for the feedback, and for pointing out the wrong error message. I have fixed that and added a test.

> Two smaller things:
>
> - The else after return is redundant — gofmt/staticcheck won't flag it, but a plain second return reads better.
> - The configdir.go change is fine, but we just landed https://github.com/gammons/slk/pull/128 which refactored this exact function to a candidate list. If you rebase, case "windows": becomes another candidate line, which fits your two-branch logic naturally — worth aligning to that shape.

Additionally, I applied these two recommendations.

gammons:
This is in. The rework addresses everything:

- `copyToTemp` now returns `ErrCookieLocked` when `isLockError` fires, so the "close Slack and try again" message at `onboarding.go:144` is actually reachable — that was the one-word gap.
- Injecting `isLockError func(error) bool` as a parameter makes it testable, and `cookiedb_test.go` now asserts a lock surfaces as `ErrCookieLocked` at that boundary — which is exactly the test that would have caught the original bug.
- The `configdir.go` windows case is now a candidate list matching the flatpak/snap shape from #128.

`go build` and `go test ./internal/slackdesktop/...` pass. CI approved and running.

One small style note (non-blocking): `var candidates = [...]string` works, but a plain slice literal `[]string{...}` with `append` is the idiom the rest of the file uses for the linux candidates. Not worth churning for.

Thanks for the quick turnaround — merging.

Allison-E:
> One small style note (non-blocking): var candidates = [...]string works, but a plain slice literal []string{...} with append is the idiom the rest of the file uses for the linux candidates. Not worth churning for.

@gammons Just fixed this formatting issue which caused the lint CI check to fail.

Issue #159, opened 2026-08-25
Mark new messages read on Slack when they arrive in the active channel


bjornsnoen:
## Problem

When a new message arrives in the channel currently being viewed, slk treats it as read locally but does not advance Slack's server-side read cursor.

The active-channel path in `internal/ui/reducer_send.go` avoids applying a local unread indicator. However, `conversations.mark` is only called when entering or loading a channel. Its timestamp therefore covers messages that existed when the channel was opened, not messages received afterward.

This can leave slk showing a message as read while Slack still considers it unread. Besides inconsistent cross-client read state, this may allow Slack to send a mobile notification even though the message is visible in slk.

Selecting a channel inside slk does not necessarily mean the user can currently see it. For example, slk may be running in an inactive terminal, tmux window, or tmux pane while the same channel remains selected.

## Expected behavior

When a top-level message or broadcast thread reply arrives in the active channel of the active workspace, slk should advance Slack's read cursor only if the terminal containing slk is currently focused.

If slk is running in an inactive terminal, tmux window, or tmux pane, the message should remain unread both locally and on Slack.

Plain thread replies should continue to use thread-specific read handling and should not advance the parent channel cursor unless broadcast.

## Implementation considerations

- Track terminal focus using Bubble Tea's `tea.View.ReportFocus` and `tea.FocusMsg` / `tea.BlurMsg`.
- Require active workspace, active channel, and terminal focus before automatically marking an incoming message read.
- Document that tmux users need `set -g focus-events on`.
- If focus state is unavailable or unknown, do not automatically mark the message read.
- Send `conversations.mark` for eligible messages received in the focused channel.
- Serialize or coalesce updates to avoid issuing an unbounded request for every message in a burst.
- Check the HTTP status and Slack JSON response before treating the mark as successful.
- Only update the local persisted read state after Slack accepts the marker.
- Log failures while leaving the message unread locally so the state can be reconciled later.

## Relevant code

- `cmd/slk/main.go`: `rtmEventHandler.OnMessage`
- `cmd/slk/main.go`: `markChannelReadAsync`
- `internal/ui/app.go`: `App.Update` and `App.View`
- `internal/ui/reducer_send.go`: `reduceNewMessage`
- `internal/slack/client.go`: `markChannel` / `MarkChannel`
- `wiki/Terminal-Compatibility.md`: tmux focus-event setup

PR #201, opened 2026-09-10
Show direct-mention counts in the sidebar

gammons:
Distinguishes an unread channel from one where you were @-mentioned, the way the official client does.

A channel with unread direct mentions renders a themed pill with the count in place of its unread dot. DMs and group DMs badge on every unread message; channels badge only on an explicit `@you`, `@here`, `@channel` or `@everyone`. Collapsed section headers carry both figures side by side — `•5` channels unread, `▐3▌` mentions.

```
  # general      ●
  # deploys     ▐3▌
  @ alice      ▐12▌
  # noisy     ▐99+▌
```

## How it works

Slack's `client.counts` already returns `mention_count` per conversation, and slk already fetched it and threw it away. That value is now persisted to a new `channels.mention_count` column and read by the sidebar through the `readStateReader` closure that already delivers read state on every render — so no new plumbing crosses the `internal/ui` networking boundary.

`client.counts` is only called at boot and reconnect, though, so a long-lived session would never see a badge change. Incoming `message` events therefore increment the count locally, and `*_marked` events overwrite it with the server's authoritative value.

The error term is what makes this shape worth it: local detection can only *undercount*, only on `@usergroup` mentions (resolving those needs your own group memberships, which slk does not have), and the error is erased the moment you read the channel or reconnect. The dot still appears regardless, so no message is ever missed outright — worst case you see a dot where Slack would have shown a `1`.

## Design decisions worth knowing

**Mentions pierce mute.** A muted channel stays grey and unbolded for ordinary traffic but still badges an explicit mention. That escape hatch is what makes muting safe on a busy channel, and it holds on rows and collapsed headers alike.

**The badge shares the dot's gate, structurally.** The increment is nested inside the existing `has_unread` write rather than repeating its conditions, so the two can never disagree about whether a message arrived unread. Everything that gate excludes — non-broadcast thread replies, self-authored messages, edits — is excluded from the badge for free.

**`app` counts as a DM.** slk classifies a bot DM as type `app` purely so the sidebar can group Apps separately; Slack reports human and app DMs alike in its `ims` block. Omitting `app` would badge an app DM from the server at boot and then never increment it live.

**Two counts stay separate on collapsed headers.** `•N` counts channels with unreads; the pill sums mentions. Merging them would answer a different question with the same glyph.

## Two unverified assumptions

Neither `mention_count` on the `ims` block of `client.counts` nor `mention_count` on `*_marked` events has been confirmed against a live capture — this repo has no fixture proving either. Both are consumed as specified and both degrade to `0` if absent, which undercounts and self-corrects at the next boot. Every code site that relies on them says so. `debuglog.WS` now logs `mention_count=`, so a live session settles it.

## Incidental cleanup

- `UnreadInfo.Count` fabricated its value — flooring to `1` on any unread channel and hardcoding `1` for every IM. Nothing read it, so the fabrication was invisible; wiring it to a renderer would have painted a `1` badge on every unread channel. Renamed to `MentionCount` and the floor removed.
- Dropped the dead `channels.unread_count` column from `CREATE TABLE`. No destructive migration — existing databases keep the vestigial column harmlessly.
- Extracted the mention predicate out of `notify.ShouldNotify` into `internal/mention` so the notification policy and the badge share one definition. `notify`'s existing tests pass unmodified, which is the regression proof.

## Two pre-existing bugs found, recorded, not fixed

Both are documented at the end of the design doc rather than fixed here, per AGENTS.md:

- The sidebar's inner render cache (`m.cacheValid`) is not keyed on `styles.Version()`, so a runtime theme change returns stale bytes. Affects every sidebar glyph.
- Section headers wrap instead of truncating — they never got the `truncate.StringWithTail` treatment channel rows have. This feature consumes up to 6 columns of headroom, so it makes the bug easier to hit without causing it.

## Merge note

This branch was cut from a `main` that went ~90 commits stale, and focus-gated read marking (#159) rewrote the code this feature nests inside. `origin/main` is merged in, resolved deliberately rather than textually:

- The mention-clear moved into `markChannelRead`, behind the early return, so a badge is never cleared for a mark Slack rejected. That guarantee did not exist before.
- The increment dropped its own self-author check, which now duplicates upstream's `isSelfMessage` inside the very gate it nests in.
- `TestOnMessage_MentionIncrementsCount`'s active-channel case flipped from `want 0` to `want 1`. Upstream deliberately removed active-channel suppression — "this channel is selected" never meant "the user can see it" — and the badge follows `has_unread` rather than re-deriving a visibility rule of its own.

## Still outstanding

- Nobody has looked at the rendered badge in a real terminal. Tests pin that the digits are contiguous and that the badge's coloured cells form exactly one run, but not whether it reads well at a narrow width.
- The two payload assumptions above.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l cmd internal` empty, `go test ./... -race` green.

Design: `docs/superpowers/specs/2026-09-09-mention-badges-design.md`
Plan: `docs/superpowers/plans/2026-09-09-mention-badges.md`


gammons:
Follow-up from testing, pushed in `c5b2f44`.

**Marking a direct mention unread showed a bare dot instead of the count.** The `u`-key path wrote `0` and relied on Slack's echoed `*_marked` event to restore the real number. That echo carries no usable `mention_count`, so the badge the user was deliberately preserving disappeared at the moment they tried to preserve it.

It now recounts from the cache: `GetMessagesSince` returns the main-feed messages at or after the new boundary, and `countMentionsSince` counts the ones that mention the user, excluding self-authored messages to match the live path's `isSelfMessage`. That is not the invented number the original comment was avoiding — the user picks the boundary off their own screen, so those messages are cached by construction. It can undercount if the boundary is far enough back that the intervening messages were never fetched, and `client.counts` corrects that on the next reconnect.

This is also partial evidence on one of the two open payload questions: `*_marked` does not appear to carry a usable `mention_count`.

**Muted channels showed mentions at full strength.** `MutedMentionBadgeStyle()` mixes the badge background toward the sidebar background, keeping the hue and dropping the brightness — the same treatment `ChannelMuted` already gives a row's foreground. Mentions still pierce mute; they just stop shouting. A collapsed section header dims its aggregate only when *every* mention it counted came from a muted channel, so one unmuted mention still makes the header shout.

Found while implementing: `mixColors` panics on a nil `color.Color`, and `SelectionBackground` is nil until `Apply()` runs — the state every sidebar test renders in, and the state at process start before a theme loads. `MutedMentionBadgeStyle` degrades to the undimmed background there rather than panicking.

The `dm`/`group_dm`/`app` rule now has two live call sites, so it is extracted to `messageMentionsSelf` rather than duplicated.

PR #202, opened 2026-09-10
fix(ui): ignore lock-state modifier bits in ctrl-key checks

laraibg786:
On terminals implementing the [Kitty Keyboard Protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), `NumLock`/`CapsLock`/`ScrollLock` state is reported as extra bits riding along in the key event's `Mod` field — even for bindings that have nothing to do with those locks. Every ctrl-key check in this codebase used exact equality (`k.Mod == tea.ModCtrl`), so on those terminals, having NumLock or CapsLock **on** silently broke every affected Ctrl-binding below.

Terminals that don't implement the protocol never set those bits, so they were unaffected, this bug is Kitty-protocol-specific.

## Fix

Added `isCtrl(mod)` in `internal/ui/compose/model.go`, which masks out `ModCapsLock | ModNumLock | ModScrollLock` before comparing to `ModCtrl`, and applied the same masking inline in `internal/ui/mode_insert.go` by stripping those bits from `mod` once, up front, before any comparison uses it. Because every check in `mode_insert.go` (Ctrl+U, Ctrl+V, Ctrl+J-as-newline, and the plain Up/Down jump-to-edge check) reads that same masked `mod`, the one change fixes all of them — the Up/Down fix isn't a separate code change, just a side effect of masking at the source. Both are no-ops on terminals that never set those bits, so behavior on non-Kitty terminals is unchanged.

## Behavior matrix

| Keybind | Terminal | Lock keys (NumLock/CapsLock) | Expected | Before fix | After fix |
|---|---|:-:|---|---|---|
| Ctrl+N/P — mention picker (`@`) | Non-Kitty | off / on | move selection | ✅ correct | ✅ correct |
| Ctrl+N/P — mention picker (`@`) | Kitty-protocol | off | move selection | ✅ correct | ✅ correct |
| Ctrl+N/P — mention picker (`@`) | Kitty-protocol | **on** | move selection | ❌ falls through to `default:`, **closes the picker** | ✅ correct |
| Ctrl+N/P — channel picker (`#`) | Non-Kitty | off / on | move selection | ✅ correct | ✅ correct |
| Ctrl+N/P — channel picker (`#`) | Kitty-protocol | off | move selection | ✅ correct | ✅ correct |
| Ctrl+N/P — channel picker (`#`) | Kitty-protocol | **on** | move selection | ❌ falls through to `default:`, **closes the picker** | ✅ correct |
| Ctrl+N/P — emoji picker (`:`) | Non-Kitty | off / on | move selection | ✅ correct | ✅ correct |
| Ctrl+N/P — emoji picker (`:`) | Kitty-protocol | off | move selection | ✅ correct | ✅ correct |
| Ctrl+N/P — emoji picker (`:`) | Kitty-protocol | **on** | move selection | ❌ falls through to `default:`, **closes the picker** | ✅ correct |
| Ctrl+U — clear compose | Non-Kitty | off / on | clear text + attachments + uploading flag | ✅ correct | ✅ correct |
| Ctrl+U — clear compose | Kitty-protocol | off | clear text + attachments + uploading flag | ✅ correct | ✅ correct |
| Ctrl+U — clear compose | Kitty-protocol | **on** | clear text + attachments + uploading flag | ❌ falls through to textarea's default binding (**delete to line start only**) | ✅ correct |
| Ctrl+V — paste | Non-Kitty | off / on | `smartPaste` (clipboard, attachment-aware) | ✅ correct | ✅ correct |
| Ctrl+V — paste | Kitty-protocol | off | `smartPaste` | ✅ correct | ✅ correct |
| Ctrl+V — paste | Kitty-protocol | **on** | `smartPaste` | ❌ falls through to textarea's default paste binding | ✅ correct |
| Up/Down at first/last line — jump to start/end | Non-Kitty | off / on | jump to start/end of textarea | ✅ correct | ✅ correct |
| Up/Down at first/last line — jump to start/end | Kitty-protocol | off | jump to start/end of textarea | ✅ correct | ✅ correct |
| Up/Down at first/last line — jump to start/end | Kitty-protocol | **on** | jump to start/end of textarea | ❌ falls through to the textarea's own cursor movement (line-relative only, never reaches column 0 / the true end) | ✅ correct |

The bug only ever manifests in the bolded **on** rows — a Kitty-protocol terminal with a lock key active. Everything else was already correct and stays correct.

## Is my terminal affected?

Only terminals implementing the Kitty Keyboard Protocol are affected, and only while NumLock/CapsLock is on. Quick check from a real TTY:

```bash
old_stty=$(stty -g)
stty raw -echo min 0 time 3        # raw mode, 0.3s read timeout
printf '\e[?u' > /dev/tty          # ask the terminal to report its flags
reply=$(dd bs=1 count=32 2>/dev/null < /dev/tty)
stty "$old_stty"

if [[ "$reply" == *'?'*u* ]]; then
  echo "Kitty keyboard protocol: SUPPORTED"
else
  echo "Kitty keyboard protocol: not supported (no reply)"
fi
```

A supporting terminal replies `ESC [ ? <flags> u`; a non-supporting one stays silent and the read times out.

Note: **Konsole does not implement this protocol**, so it is unaffected regardless of NumLock/CapsLock state.

## Reproduction (for review)

1. Open a Kitty-protocol terminal (e.g. kitty, WezTerm, foot, Ghostty).
2. Turn NumLock **on**.
3. Type `@` to open the mention picker, then press `Ctrl+N` — on `main` this closes the picker instead of moving the selection; on this branch it navigates correctly. Same for `Ctrl+P`, and for `#`/`:` pickers.
4. With some draft text in the compose box, press `Ctrl+U` — on `main` this only deletes to the start of the line; on this branch it fully clears the compose box (text + attachments).
5. Copy some text externally, then press `Ctrl+V` in the compose box, on `main` this falls through to the textarea's default paste; on this branch it goes through `smartPaste`.
6. Type a multi-line draft, put the cursor on the first line, and press `Up` — on `main` this just moves within the line instead of jumping to column 0 of the draft; on this branch it jumps to the true start. Same for `Down` on the last line jumping to the true end.

## Test plan

- [x] `TestIsCtrl_IgnoresLockStateBits` — unit coverage of the new helper across all three lock bits, alone and combined, plus a Ctrl+Shift negative case.
- [x] `TestMentionNavigateUpDown_WithNumLock` — regression test for the picker-closing bug.
- [x] `TestHandleInsertMode_CtrlUIgnoresNumLock` — regression test for Ctrl+U.
- [x] `TestHandleInsertMode_PasteIgnoresCapsLock` — regression test for Ctrl+V.
- [x] `TestInsertModeKeys/up_on_the_first_line_with_NumLock_still_jumps_to_the_start_of_the_textarea` and its CapsLock/Down counterpart — regression tests for the jump-to-edge bug.

### DIFF
--- a/internal/ui/compose/model.go
+++ b/internal/ui/compose/model.go
@@ -542,15 +542,22 @@ func (m *Model) autoGrow() {
 	}
 }
 
+// isCtrl reports whether mod represents Ctrl held, ignoring lock-state
+// bits (NumLock/CapsLock/ScrollLock) that ride along on terminals
+// implementing the Kitty Keyboard Protocol.
+func isCtrl(mod tea.KeyMod) bool {
+	return mod&^(tea.ModCapsLock|tea.ModNumLock|tea.ModScrollLock) == tea.ModCtrl
+}
+
 // handleMentionKey processes key events when the mention picker is active.
 func (m Model) handleMentionKey(msg tea.KeyMsg) (Model, tea.Cmd) {
 	k := msg.Key()
 	switch {
-	case k.Code == tea.KeyUp || (k.Code == 'p' && k.Mod == tea.ModCtrl):
+	case k.Code == tea.KeyUp || (k.Code == 'p' && isCtrl(k.Mod)):
 		m.mentionPicker.MoveUp()
 		return m, nil
 
-	case k.Code == tea.KeyDown || (k.Code == 'n' && k.Mod == tea.ModCtrl):
+	case k.Code == tea.KeyDown || (k.Code == 'n' && isCtrl(k.Mod)):
 		m.mentionPicker.MoveDown()
 		return m, nil
 
@@ -639,11 +646,11 @@ func (m *Model) insertMention(result *mentionpicker.MentionResult) {
 func (m Model) handleChannelKey(msg tea.KeyMsg) (Model, tea.Cmd) {
 	k := msg.Key()
 	switch {
-	case k.Code == tea.KeyUp || (k.Code == 'p' && k.Mod == tea.ModCtrl):
+	case k.Code == tea.KeyUp || (k.Code == 'p' && isCtrl(k.Mod)):
 		m.channelPicker.MoveUp()
 		return m, nil
 
-	case k.Code == tea.KeyDown || (k.Code == 'n' && k.Mod == tea.ModCtrl):
+	case k.Code == tea.KeyDown || (k.Code == 'n' && isCtrl(k.Mod)):
 		m.channelPicker.MoveDown()
 		return m, nil
 
@@ -1121,11 +1128,11 @@ func (m *Model) maybeOpenEmojiPicker() {
 func (m Model) handleEmojiKey(msg tea.KeyMsg) (Model, tea.Cmd) {
 	k := msg.Key()
 	switch {
-	case k.Code == tea.KeyUp || (k.Code == 'p' && k.Mod == tea.ModCtrl):
+	case k.Code == tea.KeyUp || (k.Code == 'p' && isCtrl(k.Mod)):
 		m.emojiPicker.MoveUp()
 		return m, nil
 
-	case k.Code == tea.KeyDown || (k.Code == 'n' && k.Mod == tea.ModCtrl):
+	case k.Code == tea.KeyDown || (k.Code == 'n' && isCtrl(k.Mod)):
 		m.emojiPicker.MoveDown()
 		return m, nil
 
--- a/internal/ui/compose/model_test.go
+++ b/internal/ui/compose/model_test.go
@@ -294,6 +294,65 @@ func TestMentionNavigateUpDown(t *testing.T) {
 	}
 }
 
+func TestIsCtrl_IgnoresLockStateBits(t *testing.T) {
+	cases := []struct {
+		name string
+		mod  tea.KeyMod
+		want bool
+	}{
+		{"plain ctrl", tea.ModCtrl, true},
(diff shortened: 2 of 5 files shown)


gammons:
This is a real Kitty-protocol bug and the behavior matrix makes it easy to verify the scope. I checked the two things that matter:

**The fix is complete.** I grepped every exact-equality `Mod` comparison in the codebase. `mode_insert.go` and `compose/model.go` are the only two places doing raw `== tea.ModCtrl` — and both are covered. Every other mode goes through `key.Matches`, which compares the normalized key string and isn't affected. There's no third file left broken.

**The test has teeth.** I removed the mask from `mode_insert.go:116` and `TestHandleInsertMode_PasteIgnoresCapsLock` fails:

```
--- FAIL: TestHandleInsertMode_PasteIgnoresCapsLock
```

Masking at the source (`mod := msg.Key().Mod &^ ...` on line 116) is the right call — it means the `Ctrl+V`, `Ctrl+U`, and Up/Down checks below all inherit the fix rather than each needing its own change, which is exactly why the Up/Down fix "isn't a separate code change." And masking only the three lock bits is a no-op on non-Kitty terminals, so nothing else shifts.

Build, vet, gofmt clean; the insert/compose suites pass under `-race`.

One note for the description: the `isCtrl` helper in `compose/model.go:549` and the inline mask in `mode_insert.go` are two spellings of the same predicate. Not worth churning now, but if a third site ever appears, that's the moment to lift a shared helper into one place.

Merging.

PR #203, opened 2026-09-10
fix(ui): keep the theme background on Block Kit lines so selection tints them

rfist:
## Problem

Selected Block Kit messages looked fragmented: the row padding had the selection tint, but the text still used the terminal's default background.

Regular message text gets its background from `styles.MessageText`. Block Kit lines do not have that outer style. When an avatar is present, `placeAvatarBeside` prepends the gutter after the Block Kit content is rendered; the gutter's closing ANSI reset then clears the background for the rest of the line. Inline bold, italic, link, and mention spans add more resets with the same effect.

The avatar matters. Without it, the first run on each line does not encounter the gutter reset and the bug is easy to miss in tests.

## Fix

`WithBackground` makes each Block Kit line carry its own background:

- prefix the line with the current background, covering the run after the avatar gutter;
- reapply that background after ANSI resets inside the line.

Both message panes use the helper when joining Block Kit and legacy-attachment lines:

- `renderMessagePlain` in `internal/ui/messages/model.go`;
- `renderThreadMessage` in `internal/ui/thread/model.go`.

Only the background is restored. Kitty placeholders encode their image ID in the foreground, so changing the foreground could make the terminal reference the wrong image.

The helper uses the existing `ReapplyBgAfterResets` path. This PR also makes that function honor its documented contract for both reset forms: Lip Gloss's `ESC[m` and the explicit `ESC[0m` emitted by other ANSI producers.

## Tests

There is one regression test for each production call site:

- `TestBuildCache_BlockKitRunsKeepBackground`;
- `TestRenderThreadMessage_BlockKitRunsKeepBackground`.

Both tests render through the real message path with a stand-in avatar. Reverting either call site makes its corresponding test fail with `Block Kit run rendered with no background`.

Helper-level tests cover the line prefix, both ANSI reset forms, empty-background passthrough, and the detector used by the assertions. These tests supplement the call-site tests; they do not replace them.

The seven affected golden snapshots were regenerated. Their visible text is unchanged; only the background SGR at the start of Block Kit lines differs.

## Test plan

- Reverted both production call sites and confirmed the two call-site tests fail.
- Confirmed the explicit `ESC[0m` case fails before the reset handling fix and passes afterwards.
- `gofmt -l .`
- `go vet ./...`
- `go test -count=1 ./...`
- `go test -race ./...`
- `git diff --check` outside the golden ANSI fixtures, whose trailing spaces are part of the snapshots
- Rebased onto current `main`.


gammons:
Verified against the golden files, which is the right way to check this — the `thread_open.ansi` diff shows Block Kit lines now carrying `48;2;26;26;46` (the theme background) through the inline resets, where before they reverted to the default background mid-line.

The avatar observation is the part that makes this a real find rather than a tidy-up: the bug only manifests when the avatar gutter's closing reset is present to clobber the background, so tests that render a Block Kit message with no avatar would see a clean line and never catch it. You've got a regression test per call site (`renderMessagePlain` and `renderThreadMessage`) and the `ReapplyBgAfterResets` fix handles both `ESC[m` and `ESC[0m`, which is the detail that makes it correct for non-Lipgloss producers.

Leaving the foreground alone is right — the kitty image-ID encoding is fragile enough without touching it.

Build, vet, gofmt clean; 55 packages green under `-race` locally. Merging.

PR #204, opened 2026-09-10
fix(messages): keep fenced code blocks intact through the message body wrap

rfist:
## Problem

Fenced code blocks rendered as ragged fragments instead of a solid surface. Leading indentation disappeared, internal spacing could collapse, and each line's background stopped at its last character.

Two independent behaviors caused it:

1. `codeBlockStyle` had no width, so short lines were not padded to the message width.
2. `WordWrap` passed every line through `strings.Fields`, even when the line already fitted. That reflow removes indentation and collapses whitespace.

The second change affects every message line, not only code blocks. The new behavior is deliberately narrow: fitting lines are preserved byte-for-byte, while overlong prose still reflows at word boundaries.

## Fix

The message and thread panes now pass their content width through `RenderSlackMarkdownOpts`.

For fenced code blocks, the renderer:

- removes only the line break next to each fence, preserving indentation and trailing whitespace in the code itself;
- hard-wraps code to `Width-2`, leaving one padding column on each side;
- renders the block at the full content width so its surface background reaches the right edge.

Code is hard-wrapped rather than word-wrapped because moving a token to the next line changes the pasted content.

`WordWrap` now returns any fitting line unchanged. Its existing `strings.Fields` path remains in place for lines that actually exceed the limit.

`Width: 0` preserves the previous rendering path for callers such as compact previews that do not provide a content width.

## Tests

`internal/ui/messages/codeblock_wrap_test.go` covers the observable contracts:

- every code-block line fills the available width;
- indentation on the first code line survives;
- an unbroken long line stays within the width;
- fitting prose retains its exact whitespace;
- overlong prose still wraps at word boundaries;
- a zero-width options value still renders the code block.

The regression checks were run against each production change separately:

- removing the fitting-line guard fails the full-width, indentation, and fitting-line tests;
- removing the code-block width branch fails the full-width test;
- restoring the old `strings.TrimSpace` call fails the first-line indentation test.

Golden snapshots are unchanged.

## Test plan

- `gofmt -l .`
- `go vet ./...`
- `go test -count=1 ./...`
- `go test -race ./...`
- `git diff --check`
- Rebased onto current `main`.

### DIFF
--- a/internal/ui/messages/codeblock_wrap_test.go
+++ b/internal/ui/messages/codeblock_wrap_test.go
@@ -1,0 +1,95 @@
+// Fenced code blocks must retain their whitespace and full-width surface
+// background when the message body is wrapped.
+package messages
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/charmbracelet/x/ansi"
+)
+
+// renderBody runs the same two steps the message panes run: render the
+// markdown, then wrap the result to the content width.
+func renderBody(t *testing.T, src string, width int) []string {
+	t.Helper()
+	out := RenderSlackMarkdownWith(src, RenderSlackMarkdownOpts{Width: width})
+	return strings.Split(WordWrap(out, width), "\n")
+}
+
+func TestCodeBlock_EveryLineFillsTheFullWidth(t *testing.T) {
+	const width = 40
+	lines := renderBody(t, "```\nshort\nalso short\n```", width)
+
+	var body []string
+	for _, l := range lines {
+		if ansi.Strip(l) != "" {
+			body = append(body, l)
+		}
+	}
+	if len(body) != 2 {
+		t.Fatalf("got %d content lines, want 2:\n%q", len(body), lines)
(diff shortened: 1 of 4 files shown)


gammons:
Two independent bugs correctly separated, and the tests have teeth — I removed the fitting-line early return in `wrapLine` and watched both `TestCodeBlock_EveryLineFillsTheFullWidth` and `TestCodeBlock_PreservesIndentation` fail.

The `WordWrap` change is the load-bearing one and it's well-scoped: fitting lines pass through byte-for-byte, overlong prose still reflows via `strings.Fields`. That preserves the code-indentation fix without touching the soft-wrap behavior everything else depends on. And hard-wrapping code at `Width-2` rather than word-wrapping it is the right call — moving a token changes what the user copied.

`Width: 0` keeping the old path for compact previews is a good escape hatch.

Build, vet, gofmt clean; 55 packages green under `-race`. Merging.

PR #205, opened 2026-09-11
feat(compose): support external editors using ctrl + e

### LINKED ISSUE #200: Support editing the message in external editors in addition to built-in compose box
## Motivation

The compose box works fine for quick one-line messages, but it's a fairly basic text field. It doesn't support the kind of editing power a lot of terminal users are used to from their own editor, word-by-word movement, undo history, visual/selection-based editing, copy-paste between other things they have open, search-and-replace, snippets, and so on. Vim users in particular are used to being able to jump around and edit text quickly, and none of that is available today when writing a Slack message in slk.

This is the same problem `git commit`, `crontab -e`, and many other CLI tools solved a long time ago: rather than building a full text editor into the tool itself, they just hand the text off to whatever editor the user already has configured (`$EDITOR`/`$VISUAL`) and read it back once the user is done.

## What this would let a user do

- Compose or edit a message using their own editor of choice (vim, neovim, helix, nano, VS Code, whatever `$EDITOR`/`$VISUAL` points to), with all of that editor's own features available.
- Do this for both new messages and thread replies.
- Come back to slk afterward with the message ready to review and send, without slk needing to know anything about the editor itself.
- Ideally this should make the compose box read-only to avoid conflicts while listening to the write/modify events from OS and update the contents of the read-only box. the compose box should read from the file and become active once the editor exists.

laraibg786:
Writes the compose box to a temp file and suspends the TUI to edit it externally, reloading the result on exit. Editor is resolved once at startup ($VISUAL, then $EDITOR, then compose.editor in config.toml) and stored on App rather than re-read per keypress. Compose locks read-only while the editor is open. when there is some issue opening buffer externally, there is toast shown to the user rather than silent skip.

closes #200

### DIFF
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -1174,6 +1174,9 @@ func run() error {
 	app.SetSidebarStaleThreshold(time.Duration(cfg.Sidebar.HideInactiveAfterDays) * 24 * time.Hour)
 	app.SetMouseWheelLines(cfg.Appearance.MouseWheelLines)
 	app.SetColoredUsernames(cfg.Appearance.ColoredUsernames)
+	if editor, ok := ui.ResolveEditor(cfg.Compose.Editor); ok {
+		app.SetComposeEditor(editor)
+	}
 
 	// Wire theme switcher
 	app.SetThemeItems(styles.ThemeNames())
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -19,9 +19,15 @@ type Config struct {
 	Notifications Notifications         `toml:"notifications"`
 	Cache         CacheConfig           `toml:"cache"`
 	Sidebar       Sidebar               `toml:"sidebar"`
+	Compose       Compose               `toml:"compose"`
 	Sections      map[string]SectionDef `toml:"sections"`
 	Theme         Theme                 `toml:"theme"`
 	Workspaces    map[string]Workspace  `toml:"workspaces"`
+}
+
+type Compose struct {
+	// Editor is the Ctrl+E fallback when $VISUAL/$EDITOR are unset.
+	Editor string `toml:"editor"`
 }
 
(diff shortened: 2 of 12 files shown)


gammons:
This is well built and I'm merging it. Closes #200.

A few things that show you actually ran it rather than reasoned about it:

- The `editorCommand` comment about real-tty-vs-pipe — `tea.ExecProcess` falls back to a non-`*os.File` writer, which forces a pipe through a copy goroutine, which breaks the editor's terminal negotiation. Setting all three of Stdin/Stdout/Stderr to the real fds is the fix, and it's the kind of detail that's invisible until it isn't.
- Trusting the file's contents over the editor's exit code (`launchFailed := m.Err != nil && !errors.As(m.Err, &exitErr)`) is correct — some editors exit non-zero on a saved draft, and treating that as "never opened" would eat the user's text. You only toast a failure when the editor genuinely never launched.
- Compose locks read-only while the editor is open (`editingExternally` gates `Update`), and the placeholder changes to say what's happening, so the UI isn't silently dead.
- The temp file is cleaned up on every path, including the `os.Remove` in `reduceEditorFinished`'s defer.
- Editor is resolved once at startup (`$VISUAL`, `$EDITOR`, then `compose.editor`) and stored on `App`, not re-read per keypress — and no `vi` fallback, since it isn't guaranteed on Windows.

You also rebased onto the #202 lock-bit masking, so `Ctrl+E` inherits the NumLock/CapsLock fix automatically.

Docs are updated in three places (Configuration, Features, Keybindings) and there's a 282-line test file.

One follow-up, not blocking: when the editor exits with unsaved content the draft is whatever's in the file, which is the right call — but if the user `dd`'d everything and saved, they get an empty compose with no indication. A "draft emptied by editor" toast might be worth it. Your call.

Merging.

PR #207, opened 2026-09-11
fix(rail): make the workspace unread dot agree with the sidebar

landonforshage:
## Problem

The green dot next to a workspace in the left rail lit up for any channel with `has_unread=1` in the cache. Every other unread indicator in slk (the sidebar dot, the `(N)` in the window title, `$SLK_UNREAD`) ignores muted channels. The rail didn't. So a single muted firehose channel kept a workspace's dot on permanently, with nothing visibly unread inside it.

Worse, one workspace was lit by a channel that wasn't in the sidebar or Ctrl+t at all, so there was no way to clear it from inside slk. That one turned out to be an archived Slack Connect channel.

I hit both within an hour of first use, on 4 of 5 workspaces. Once the dot is always on, you stop looking at it.

## Cause

Two things, verified against live `client.userBoot` and `client.counts` responses on 2026-09-11:

1. **Mutes aren't in the database.** `WorkspacesWithUnreads` was a plain SQL query. Mute state lives only in `service.MuteStore`, in memory, so no query can filter it.
2. **The cache knows about channels the sidebar hides.** `client.userBoot` lists archived channels the user belongs to with `is_archived: true`. `bootConversations` correctly drops them from the sidebar, but `hydrateFirstSight` caches every channel userBoot returns, archived or not. `client.counts` still reported the archived channel as unread (its `last_read` was Slack's never-opened sentinel), so the row got `has_unread=1` and nothing could ever show or clear it. Channel rows are never deleted, so the dot was permanent.

## Fix

The rail now asks the same question the sidebar asks: *does this workspace have a channel in its list that `IsVisiblyUnread`?*

- `cache.UnreadChannels()` returns unread rows **with their channel IDs** (replaces `WorkspacesWithUnreads`).
- `railUnreadWorkspaces` (`cmd/slk/rail_unread.go`) is a pure function: for each unread row, find the channel in the workspace's `wctx.Channels` and apply `IsVisiblyUnread`. Muted → no dot. Not in the list → no dot. Workspace still connecting → keep the dot (the conservative default `MuteStore.Ready` documents, and what keeps cached dots visible during boot).
- The title's `+N` and `$SLK_OTHER_UNREAD` read through the same function, so they stay in sync with the dots.

Review of the diff turned up three more gaps, each its own commit:

- **`workspaceRouter` gets a mutex.** Its map was written by connect goroutines and read by UI callbacks with no lock; the comment saying it was populated before `p.Run` was wrong, since connects run alongside it. That's a runtime fatal waiting to happen, and the rail reader made it more likely. The duplicate `workspaces` map in `run` is removed too.
- **Inactive-workspace mute changes refresh the rail.** Muting a channel from another client in a workspace you're not looking at updates the dot immediately instead of waiting for the next unrelated event.
- **Channel-list changes fan out to the rail and title.** `SectionsRefreshedMsg` for the active workspace and `WorkspaceReadyMsg` both changed what the unread surfaces derive from without calling `notifyReadStateChanged`. Muting an unread channel in the active workspace left `(N)` stale, and a workspace finishing its connect kept a cached dot its now-known channel list no longer justified.

Six commits, one per change (the fifth only trims a comment), each with its own tests.

## Why not fix it elsewhere

- **Filter archived channels out of `hydrateFirstSight`?** The cache is allowed to know more than the sidebar shows (search results resolve `<#C…>` mentions through those rows), and it wouldn't fix existing caches since rows are never deleted. Happy to add it as a follow-up if you'd like the rows gone too.
- **Add an `is_archived` column?** A migration for a display bug. Not needed once the rail checks the live list.
- **Filter at the counts write path?** Three writers to keep in sync instead of one reader.

## Tests

All new tests are plain `testing.T`, white-box. The rail tests add two unexported helpers local to their file (`unreadRow`, `railLookup`); nothing shared.

- Cache: `TestUnreadChannels`.
- Rail predicate: muted → dark, muted + unmuted → lit, absent from list → dark, unknown workspace → lit, and `MuteStore` not ready → lit (through the real `buildChannelItem`).
- `TestRailUnreadWorkspaces_ArchivedChannelInCache`: the archived-channel repro built the way production builds it (`hydrateFirstSight` → `bootConversations` → `ReplaceWorkspaceReadState`).
- `TestRailUnreadWorkspaces_RailAndTitleAgree`: dots and `OtherUnreadCount` are the same set.
- `TestWorkspaceRouter_ConcurrentAddAndByID`: fails under `-race` with the mutex removed.
- `TestMuteRefreshMsg`: active vs inactive message choice.
- `TestWorkspaceReady_RefreshesRailAndTitle` and `TestSectionsRefreshed_ActiveWorkspace_RefreshesTitle`: through `App.Update`; the title and rail recompute when a workspace connects and when an active-workspace channel is muted.

Each commit's tests were confirmed red with its change reverted. `go build`, `go vet`, `go test ./... -race`, `gofmt -l`, and `golangci-lint run` (v2.13.1) are all clean.

**Field check (macOS, five live workspaces):** with the patched build sitting on one workspace, marking a message unread in a *different* workspace from the official client lit that workspace's dot; muting the channel there cleared the dot immediately, without touching slk; unmuting lit it again; opening the channel cleared it. That is the mute rule and the inactive-workspace refresh end to end. The archived-channel case was consumed before the fix (opening it in the official client is the only way out today), so it is covered by the production-shaped test above rather than a live re-run.

## Noted but not changed

- `wctx.Channels` is read on the UI goroutine while the WebSocket handler mutates it. That predates this PR (the `Lookup` callback does the same) and needs a lock or snapshot on `WorkspaceContext`, which is bigger than this change.
- `OtherUnreadCount` will count a workspace that's in the cache but no longer configured. Pre-existing.

## AI assistance

Written with Claude (Fable 5.1, extra-high thinking effort) driving the investigation, code, and tests from a brief I wrote after reproducing both bugs. The cause was confirmed against live API responses, not inferred from the code, and Codex reviewed the diff; its findings are commits 3, 4 and 6. I've read the diff and will defend it in review.

### DIFF
--- a/cmd/slk/event_handler_test.go
+++ b/cmd/slk/event_handler_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gammons/slk/internal/cache"
 	"github.com/gammons/slk/internal/config"
+	"github.com/gammons/slk/internal/ui"
 	"github.com/gammons/slk/internal/ui/channelfinder"
 	"github.com/gammons/slk/internal/ui/sidebar"
 	"github.com/slack-go/slack"
@@ -575,5 +576,35 @@ func TestOnThreadSubscriptionChanged_PersistsOnInactiveWorkspace(t *testing.T) {
 	}
 	if len(got) != 1 || got[0].ChannelID != "C1" || !got[0].Active {
 		t.Fatalf("inactive workspace must still persist thread_subscribed; got %+v, want 1 active row", got)
+	}
+}
+
+// TestMuteRefreshMsg pins which message a mute change posts: the active
+// workspace gets its sidebar list swapped, an inactive one gets the
+// read-state signal the rail, title and status hook re-derive from.
+// The inactive case is the one that used to post nothing.
+func TestMuteRefreshMsg(t *testing.T) {
+	channels := []sidebar.ChannelItem{{ID: "C1", IsMuted: true}}
+
+	msg := muteRefreshMsg(true, "T1", channels)
+	sr, ok := msg.(ui.SectionsRefreshedMsg)
+	if !ok {
+		t.Fatalf("active: got %T, want ui.SectionsRefreshedMsg", msg)
+	}
+	if sr.TeamID != "T1" || len(sr.Channels) != 1 || !sr.Channels[0].IsMuted {
+		t.Errorf("active: msg = %+v", sr)
+	}
+	sr.Channels[0].IsMuted = false
+	if !channels[0].IsMuted {
+		t.Error("active: the message shares the handler's slice; the App would be mutating wctx.Channels")
+	}
+
+	msg = muteRefreshMsg(false, "T1", channels)
+	rs, ok := msg.(ui.ReadStateChangedMsg)
+	if !ok {
+		t.Fatalf("inactive: got %T, want ui.ReadStateChangedMsg", msg)
+	}
+	if rs.WorkspaceID != "T1" {
+		t.Errorf("inactive: WorkspaceID = %q, want T1", rs.WorkspaceID)
 	}
 }
--- a/cmd/slk/main.go
+++ b/cmd/slk/main.go
@@ -5102,14 +5102,10 @@ func (h *rtmEventHandler) refreshMutedForActive() {
 // Inactive: ReadStateChangedMsg. There is no sidebar on screen for
 // this workspace, but its rail dot, the title's "+N" and
 // $SLK_OTHER_UNREAD are all derived from these flags by
-// railUnreadWorkspaces, and nothing re-reads them until the next
-// read-state event. Before this branch existed the inactive case
-// posted nothing, so a channel muted or unmuted from another Slack
-// client left the rail stale -- an unmute left a workspace with
-// unreads dark, a mute kept one lit -- until an unrelated message or
-// mark happened to arrive. ReadStateChangedMsg is what
-// notifyReadStateChanged already answers to, and the App ignores its
-// fields, so no new message type is needed.
+// railUnreadWorkspaces, and nothing else re-reads them until the next
+// read-state event. ReadStateChangedMsg is what notifyReadStateChanged
+// already answers to, and the App ignores its fields, so no new
+// message type is needed.
 //
 // Pure so the choice is testable without a *tea.Program.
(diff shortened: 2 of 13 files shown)


gammons:
This is a real bug, correctly diagnosed, and the fix is the right shape. Merging.

**I verified the two load-bearing claims rather than trusting them:**

1. Mutes live only in `internal/service/mutestore.go` (an in-memory `map[string]bool`), so a plain SQL query structurally cannot filter them. Correct.
2. `IsArchived` is dropped from the sidebar at `bootstrap_adapters.go:408` but `hydrateFirstSight` still caches it — so the dot stays lit with no way to clear it. Correct. Your field-note about the archived Slack Connect channel explains how you hit it.

**The fix routes the rail through `IsVisiblyUnread`, the single predicate the sidebar dot, window-title count, and `$SLK_UNREAD` already share.** That is the right call. The worst version of this fix would have been a third parallel unread rule that drifts from the other two; you did the opposite.

**I checked the things that usually bite:**

- The "channel not in `wctx.Channels` → never light" branch is safe for DMs — `GetChannels` requests `im`/`mpim` and `connectWorkspace` appends them, so DM unreads still light the rail. The unlisted case is only channels the sidebar can't render (archived, left-since), where a lit dot is unexplainable.
- The `wctx == nil` → light-it default preserves boot dots and matches `MuteStore.Ready()`'s conservative assumption. Sensible.
- I reverted the production change and confirmed the tests fail — `TestRailUnreadWorkspaces_RailAndTitleAgree` and the archived-channel repro both go red as you claimed, and the archived repro is built through the real production pipeline (`hydrateFirstSight` → `bootConversations` → `buildChannelItem`), not hand-rolled fixtures. That's the part that makes it trustworthy.
- You also fixed a genuine concurrent-map fatal in the workspace router that an old comment declared impossible.

**On the `wctx.Channels` race you flagged:** your framing is right. It's a pre-existing non-fatal tear, not the fatal map race, and this PR adds a reader rather than creating the convention. I've filed #208 to track locking `WorkspaceContext.Channels` — worth doing, not worth blocking this.

Your AI-assistance note is exactly the standard I want to see: reproduced on live workspaces, confirmed the cause against real API responses rather than inferring it, and you're prepared to defend the diff. More of that, please.

Build, vet, gofmt, golangci-lint, and `go test ./... -race` all clean. Merging.

Issue #182, opened 2026-09-06
Thread header separator uses ASCII '-' where every other separator uses '─'


gammons:
## Summary

The thread pane's header separator is drawn with the ASCII hyphen `-`, while every other separator in the UI uses the box-drawing character `─` (U+2500). The thread pane's own reply separator uses `─`, so the inconsistency is visible within a single pane.

## Location

```go
// internal/ui/thread/model.go:1306
separator := lipgloss.NewStyle().
    Width(width).
    Background(styles.Background).
    Foreground(styles.Border).
    Render(strings.Repeat("-", width))
```

Compare, in the same file:

- `internal/ui/thread/model.go:1359` — `Render(strings.Repeat("─", width))`
- `internal/ui/thread/model.go:1522` — `replySeparator := separatorStyle.Render(strings.Repeat("─", width))`

## Impact

Cosmetic only. Visible whenever the thread pane is open: the header rule is a row of hyphens with visible gaps, directly above reply rules drawn as a solid line.

## Fix

Change the `"-"` at `internal/ui/thread/model.go:1306` to `"─"`.

## Note

This is now pinned by golden snapshots on the `refactor/phase0-safety-net` branch. Fixing it will fail `TestGolden/thread_open` and `TestGolden/wide`; re-bless with:

```
go test ./internal/ui -run TestGolden -update
```

and confirm the re-blessed diff shows only the separator row changing.

## Provenance

Found while building golden-image tests for the architecture refactor (Phase 0). Recorded rather than fixed, since that phase is behaviour-preserving by construction.

Issue #181, opened 2026-09-06
View() renders 6 columns wider than the terminal width, wrapping the status row


gammons:
## Summary

`App.View()` returns a frame **6 columns wider than the width it was given**. In a real terminal this overflows: the status row wraps onto an extra line, which pushes the frame down and scrolls the top row off screen.

## Evidence

Measured with `ansi.StringWidth` (OSC-safe) against the golden snapshots added in the Phase 0 test work. Every line of every frame is a single uniform width — the overflow is not ragged:

| scenario | declared | rendered width | lines |
|---|---|---|---|
| base | 120x36 | **126** | 36 |
| drag_selection | 120x36 | **126** | 36 |
| no_sidebar | 120x36 | **126** | 36 |
| thread_open | 140x30 | **146** | 30 |
| wide | 200x50 | **206** | 50 |
| narrow | 80x24 | **86** | 24 |
| window_split | 160x40 | **166** | 40 |
| overlay_finder | 120x36 | 120 | 36 |

`overlay_finder` is the only one that is correct, and only by accident: `maybeWrapFinalScreen` clamps the frame, but it is gated on `a.overlayActive()` (`internal/ui/view_overlays.go:108-111`). With no overlay open, nothing clamps.

## Root cause

`internal/ui/statusbar/model.go` builds the status row as `left + filler + rightContent + rightPad`, having budgeted the content at `width - 1`. But the spacer segments are each rendered through `styles.StatusBar`, which carries `Padding(0, 1)`:

```go
// internal/ui/styles/styles.go:467
StatusBar = lipgloss.NewStyle().
    Background(SurfaceDark).Foreground(TextPrimary).Padding(0, 1)
```

Every `styles.StatusBar.Render(...)` therefore adds 2 columns that the width budget never accounted for. In `model.go:340-357` that happens three times on the help-hint branch (`leftPad`, `rightPad`), once on the no-hint branch, and once more for the trailing `rightPad` gutter — which renders a single space as 3 columns.

There appear to be two overflow magnitudes depending on which branch runs: `+6` when the help hint is shown (all measurements above) and `+4` on the branch where the hint is dropped for lack of room. A sufficiently narrow terminal would hit the second.

## Impact

Real. Any terminal whose width equals the reported width will wrap the status row. The visible symptom is the top line of the message pane scrolling away, which reads as a rendering glitch rather than an overflow.

## Reproduction

[...]

PR #212, opened 2026-09-13
refactor(ui): decouple tui from slack I/O and core

laraibg786:
`internal/ui` no longer does I/O itself. Every call it makes to Slack, SQLite, the filesystem, the clipboard, the external editor, or the OS goes through an interface in the new `internal/core` package, and `cmd/slk` wires the implementations. A test enforces the boundary.

Nothing changes for the user. There are no bug fixes and no new features. The diff is code motion, type aliases and rewiring.

### Why
**Diagnosis**: when something breaks between the TUI and the app, the suspects are the methods on one interface, not closures spread across `app.go` and `main.go`.
**Testing**: the App can be tested with fakes for exactly the methods a test cares about, with no real filesystem, clipboard or process.
**Fewer conflicts**: new I/O has one obvious place to go, a port in `core/services.go`. `boundary_test.go` fails any change that reaches past it.

### What moved

| Port (`internal/core`) | Replaces | Implemented by |
|---|---|---|
| `ChannelService`, `MessageService`, `ThreadService`, `ReactionService`, `SearchService` | the same five interfaces, moved from `internal/ui/services.go` | `main.go` closures (unchanged bodies) |
| `FileService` | `SetUploader`, `SetFileDownloader` | upload closure, `filedl` |
| `DesktopService` | `SetClipboardReader`, `SetStatusReporter`, the `launchOS`/`os.Stat` defaults, the thread-export writer | `launchOS` (moved to `cmd/slk`), clipboard readers, `os.Stat`, `export.SaveThread` |
| `EditorService` | Ctrl+E's temp file and `exec.Cmd` | new `internal/editor` (verbatim move) |
| `PresenceService` | `SetStatusSetter`, `SetTypingSender` | `main.go` closures |
| `SettingsService` | `SetThemeSaver`, `SetWidthSaver` | `main.go` closures |
| `UnreadService` | `SetReadStateReader`, `SetWorkspaceUnreadReader` | `main.go` closures |
| `WorkspaceService` | `SetWorkspaceSwitcher` | `main.go` closure |
| `AvatarService` | `SetAvatarFunc` | avatar cache |
| `ImageFetcher` | the concrete `*image.Fetcher` field | `*image.Fetcher` |

The value types shared by the TUI and the app (`MessageItem`, `ReadState`, `Theme`, `PendingAttachment`, Block Kit types, and so on) now live in `core`. Their old names stay as type aliases, so call sites don't change.

### Commits

Each commit builds and passes the full suite on its own. They are best reviewed in order, with `git show --color-moved`.

| Commit | What it does | Size |
|---|---|---|
| `test(ui): pin the TUI's collaborator seams` | **Tests only**, passing on current `main`: 36 + 2 scenarios that drive the App the way a user would and pin what crosses each seam | +1069 |
| `refactor(core): move the data the TUI and engine share` | types into `core`, aliases left behind | ±400 |
| `refactor(core): move the five service ports` | `ui/services.go` → `core/services.go`, `tea.Msg` → `core.Msg` | ±400 |
| `refactor(ui): put the TUI's remaining collaborators behind core ports` | the new ports and setters, `main.go` rewiring | +708/−326 |
| `refactor(styles): read custom theme files through fs.FS` | `LoadCustomThemes(fsys fs.FS)` | ±10 |
| `refactor(ui): run the external editor through a core port` | Ctrl+E I/O → `internal/editor` | +197/−69 |
| `test(ui): guard the TUI/core boundary` | `boundary_test.go` | +142 |
| `test(ui): drop an unused typing-sender helper` | lint | −8 |
| `docs: point AGENTS.md at internal/core` | invariant and helper table | ±9 |

## How regressions are ruled out

- **Frozen seam tests.** `seams_test.go` and `styles/seams_test.go` land first and pass on `main`, then stay byte-identical through every later commit. The wiring they go through lives in `seams_wiring_test.go`, the only test file the refactor commits rewrite.
- **Boundary test.** On current `main` it reports 28 violations; on this branch it reports none.
- **Goldens.** All 8 frames are byte-identical; `testdata/golden` isn't in the diff.
- **Other existing tests.** Changes are mechanical only: renamed setters, the export and editor tests moving with their functions, and a wiring line where a test used to rely on the real OS default.

### Not changed on purpose

- Environment lookups (`TMUX`, `os.UserHomeDir`, `$VISUAL`/`$EDITOR`) stay in the TUI. It reads its own terminal; the boundary test allows them.
- The engine still constructs `ui` message types, and the implementations are still closures inside `run()`. Extracting them is Phase 2 work.
- `launchOS` and the clipboard readers moved verbatim and are still untested, as before.

### Manual test plan

These are the paths that changed wiring. Each should behave the same as on `main`:

- [x] Paste an image with `Ctrl+V`, then `Enter`, in a channel and in a thread: it uploads
- [x] Paste a copied file path: it attaches the file; a missing path pastes as text
- [x] `d` on a message with a file: it downloads and opens
- [x] `o` on a message with a link opens the browser; `O` / `v` preview, then `Enter`, opens the system viewer
- [x] `S` in a thread saves the Markdown export
- [x] `Ctrl+E` opens `$EDITOR`; the saved text replaces the draft; unset editor shows the toast
- [x] Presence menu: active, away, snooze, custom snooze, end DND
- [x] Theme switcher saves globally and per workspace; `[` / `]` sidebar width survives a restart
- [x] Workspace switch with `1`–`9` and by clicking the rail
- [x] Typing indicator appears in another Slack client
- [x] Avatars render; unread dots, the window title's `(N) +M` and `status_command` update
- [x] On Wayland: clipboard paste still works


laraibg786:
I'll mark this ready for review once I've finished manual testing (the checklist in the description).

This is a first step rather than a full cleanup. It moves the TUI's I/O behind interfaces in `internal/core`, so the rest of the app can be tested without the UI, and a problem between the two can be traced to one interface instead of closures scattered across `app.go` and `main.go`. Much of what's in the architecture plan is still ahead: splitting `run()`, `App.Wire(Deps)`, and so on. I've deliberately left those out to keep the blast radius small and the review manageable. The "Not changed on purpose" section lists what I skipped.

Suggestions welcome. I know this will conflict with some open PRs; #109 is the one that needs real changes, since its new service belongs in `core/services.go` now. I'm happy to rebase this after whatever you want to land first, or to help rebase the others onto it.

laraibg786:
@gammons the PR is ready from my side, i have tested the scenarios listed in the description manually, the test suite is green. please review and let me know if you want me to explain something. Your review and suggestions are highly welcome.

gammons:
Thanks for this — it's a well-built PR, and the verification story is the best part. Before getting into specifics I should give you some context you don't have.

## Context: there's a refactor plan in flight

I've been working with an agent over the last week on the parts of the architecture that hurt most. The output is on `main`:

- `docs/superpowers/plans/2026-09-06-architecture-refactor.md` — measured baseline at `4184e60`, eight findings, a six-phase sequence
- `docs/superpowers/specs/2026-09-06-phase0-test-safety-net-design.md` + its plan — Phase 0, which is what landed the goldens, the `newTestApp` harness, the mode-handler characterization and the flake fixes you're building on top of

Phase 0 is done. Phases 1–6 hadn't started, which is why this PR doesn't collide with anything. Worth reading the tracking doc if you're planning more of this — not to constrain you, but so we're not solving the same thing twice from different directions. That's the failure mode I'm most worried about right now.

## Your "Phase 2" vs the plan's Phase 2 — partial overlap

You wrote: *"the implementations are still closures inside `run()`. Extracting them is Phase 2 work."*

That maps almost exactly onto the plan's Phase 2, step 6: *extract `wireCallbacks` (541 lines) to `cmd/slk/callbacks.go` with an explicit deps struct.* Same work, same motivation.

Two things you'd hit that the plan already scoped, though:

**A prerequisite.** `p *tea.Program` is declared at `main.go:1186` and not assigned until `:1972`, so every closure in `run()` depends on the nil-then-set pattern. You can't lift `wireCallbacks` out until `p` becomes a `send func(tea.Msg)` indirection — that's step 4, and it exists specifically to unblock step 6. `newUserResolver`, `membership.New` and `resolveDMNames` already take the callback form, so the pattern is established.

**Three live data races in the same code.** Phase 2 steps 1–3 are not cleanup, they're bug fixes:

- `activeTeamID` (`main.go:1188`) is a plain `string` written from the UI goroutine (`:1906`) and from N connect goroutines (`:2035`, `:2042`), and read from every WebSocket goroutine via the `isActive` closure (`:2062`). It's also redundant with `router.Active().TeamID`.
- `workspaces` (`main.go:1187`) is a plain map written from N connect goroutines at `:2022` — and it duplicates `router.all`, written on the very next line.
- `cfg` is mutated in place by `SetThemeSaver`/`SetWidthSaver` on the UI goroutine while a copy sharing the `cfg.Workspaces` map lives in every `rtmEventHandler` (`:2070`).

The doc comment at `main.go:276-278` claims all `router.all` writes precede `p.Run`. They don't — connect goroutines start at `:2001`, `p.Run()` is at `:2186`. If you take this on, please fix the races and the comment in the same pass, and ideally before the extraction, so the extraction diff stays reviewable.

**And you've found a gap in the plan.** Your other deferral — *"the engine still constructs `ui` message types"* — isn't in any of the six phases, and after this PR it's the most visible asymmetry left: I/O now flows `ui → core ← cmd`, but messages still flow `cmd → ui` directly. `cmd/slk` imports `internal/ui` in four non-test files and constructs 73 `ui.*Msg` values. I'll add that to the tracking doc as its own item rather than smuggling it into Phase 2, since it's a different shape of problem.

One note on the net effect here: `main.go` went 5,256 → 5,270 on this branch. That's fine for what the PR set out to do, but it means the largest file in the repo got marginally larger while the I/O left the TUI. Phases 1–2 are where that gets paid back, and this PR makes them more urgent rather than less.

## You corrected something I got wrong

The invariant I had in `AGENTS.md` said `internal/ui` must not import networking and that all I/O already crossed the five service interfaces. Your boundary test finds **28 violations** on `main` — `app.go` alone imports `os/exec`, `cache`, `config`, `export`, `filedl` and `clipboard`, and calls `os.MkdirAll`, `os.WriteFile` and `os.Stat` directly; `editor.go` does five raw `os` calls; `styles/themes.go` reads the filesystem.

The original claim was narrowly true about *networking* imports in the *root* package, and I over-generalised it to "does no I/O." Your rewrite is right, and backing it with an enforcing test instead of prose is exactly the convention the file asks for. Thanks for maintaining the helper table too.

## What I verified

I didn't take the claims on trust — the regression argument is the whole review here, so:

- **38 seam tests pass on unmodified `main`.** Confirmed (I initially thought `styles/seams_test.go` didn't compile there; I'd missed the 7-line `styles/seams_wiring_test.go` shim — my error, the design is sound).
- **They stay byte-identical across all 9 commits.** Blob `a7e6abec` at every one.
- **Boundary test: exactly 28 violations on `main`, 0 here.**
- **Goldens untouched** — zero files under `testdata/golden` in the diff.
- Builds clean, `go test ./... -race` green across 56 packages, `golangci-lint` 0 issues.

The frozen-seam-test technique — land the scenarios first so they pass on `main`, freeze them, and let only a separate wiring file change — is a stronger regression argument than what the plan prescribes for the later phases. I'd like to adopt it for Phases 1–4.

## Three things I'd like addressed

**1. `TestCoreDependsOnNeitherTUINorIO` promises more than it enforces.** Its ban list is `internal/ui/*`, `charm.land/*`, `os`, `os/exec`, `net`, `net/http`, `database/sql`. It omits `internal/image`, which `core` imports directly and which itself imports `net/http`, `os` and `os/exec`. `core.ImageFetcher`'s signatures are built from `imgpkg.FetchRequest`, `FetchResult`, `Protocol`, `Render` and `KittyRenderer` (`core/services.go:894-898`), so the port layer's vocabulary is tied to an I/O package.

I don't think this blocks the PR — parallel core-owned types for the whole image pipeline is a real cost and probably not worth paying today. But a test whose name over-promises is how a boundary quietly erodes. Either add `internal/image` to the ban list and introduce core types, or narrow the test name and add a comment saying the fetcher is a deliberate exemption and why.

**2. Is `core/services.go` at 981 lines already becoming a grab-bag?** It holds 14 interfaces and 38 struct/func types. The interfaces are ports; the `*Funcs` bundles are adapters. Splitting them — ports in `core`, bundles next to their implementations, or at least separate files — would keep `core` a boundary rather than the next file we have to break up. Not a blocker, but it's the pattern this whole effort is unwinding, so I'd rather ask now.

**3. The five type aliases** (`channelfinder.Item`, `compose.PendingAttachment`, `presencemenu.Action`, `reactionpicker.EmojiEntry`, `themeswitcher.ThemeScope`) are the right call for keeping this diff mechanical. Could you open a follow-up issue to remove them? Otherwise they become permanent double-naming, and nobody will remember they were meant to be temporary.

## Also

Two production bugs surfaced during Phase 0 that are adjacent to code you touched, so you may bump into them:

- #181 — `View()` returns a frame six columns wider than the terminal, which wraps the status row. `golden_test.go` encodes it as `goldenStatusOverflow = 6`; when it's fixed that constant goes to 0 and the goldens get re-blessed.
- #182 — the thread header separator uses ASCII `-` where every other separator uses `─`.

Neither is yours to fix here.

---

Approving with the three comments above — none of them blocks merging, and I'd rather land this and iterate than hold it. Nice work on the verification discipline; it's the part I'd most like to see repeated.

laraibg786:
Thanks for your review. I have fixed your concerns and implemented your suggestions.

1. I have updated the test name to accommodate the reduced scope of the test since we are keeping the internal/image since it would be too much refactor for this PR.
2. core/services has been reduced to two ports and adopters files to make sure that this does not end up being the dump bag anyways.
3. I have opened the issue #217 to document the issue so that this can later be fixed.

This should be good for merge now.

gammons:
Looks great.  thank you @laraibg786 !

Issue #208, opened 2026-09-13
data race: WorkspaceContext.Channels read on UI goroutine, written by WS handler


gammons:
Surfaced during review of #207 (rail unread dots).

`WorkspaceContext.Channels` is a plain slice read on the bubbletea UI goroutine while the WebSocket handler goroutine writes it in place.

**Readers (UI goroutine):** `Lookup` callback in `wireCallbacks` (`cmd/slk/main.go:1435`), and after #207, `railRowLights` via `ReadStateChangedMsg` — which fires on every inactive-workspace message, so read frequency is now higher.

**Writers (WS-handler goroutine):** in-place mutation at `cmd/slk/main.go:5085` and `refreshMutedForActive`; append at `cmd/slk/main.go:4901` (`OnConversationOpened`).

**Failure mode:** torn slice header / stale bool on read — a cosmetic inconsistency, not the runtime `fatal error` that a concurrent *map* access would produce. So this is lower severity than #147's map race, but still a data race that `go test -race` will eventually flag if a test ever exercises the two concurrently.

**Fix:** put `Channels` behind an accessor with a `sync.RWMutex`, or snapshot it into an `atomic.Pointer[[]ChannelItem]` (the pattern already used for `userGroups` and `customEmoji` on the same struct — see `cmd/slk/main.go:133-143` and `:189-200` where the concurrency contract is documented). The snapshot approach fits the existing convention.

PR #215, opened 2026-09-14
fix(rail): light the workspace dot for unread thread replies

landonforshage:
## Problem

The green dot next to a workspace in the left rail has never considered thread replies. slk's own sidebar does: the Threads row carries a `•N` badge when a subscribed thread has unread replies. So a workspace can show visibly unread activity in its sidebar and a dark dot in the rail. Same class of bug as #207, one signal over.

I hit it the day after v0.20.0: two unread thread replies in a workspace I wasn't looking at, no dot. Before #207 that workspace's dot was on, but only because a muted firehose channel was lighting it by mistake; the mute rule landing is what exposed this.

## Cause

Verified against the code at `356d29e` and a live `client.counts` response on 2026-09-14:

1. **The rail reads channel rows only.** `railUnreadWorkspaces` (`cmd/slk/rail_unread.go`) took `cache.UnreadChannels()` and checked each row against `wctx.Channels`. Nothing about threads; this was true of the pre-#207 query too.
2. **A thread reply does not set the channel's `has_unread`.** `OnMessage` (`cmd/slk/main.go`, the `channelEligible` rule): only top-level messages and `thread_broadcast` replies touch the channel row. So there is no channel-side residue for the rail to pick up.
3. **Thread unread state lives in `thread_subscriptions` plus cached replies**, and the predicate the badge uses is `cache.ListSubscribedThreads` (`internal/cache/threads.go`): unread iff `max(last_read, cached_max_reply_ts, latest_reply) > last_read`, except when the newest activity is a cached reply authored by self. Slack agrees: `client.counts` for that workspace returned `threads: {has_unreads: true}` while its only channel with `has_unreads: true` was muted.
4. **The badge is active-workspace-only and in-memory** (`threadsview.Model.UnreadCount`, fed by `ThreadsListLoadedMsg`, whose reducer drops other workspaces). The rail needs the answer for every workspace, so it has to ask the DB.

## Fix

The rail now asks the same question the Threads badge answers: *does `cache.ListSubscribedThreads` return any `Unread` summary for this workspace?* Seven commits, one change each; three of them only trim or extend a comment and pin one more case:

- **`railUnreadWorkspaces` gets a thread half.** It takes the configured workspace list and a `threadsUnread(teamID, selfUserID)` input; in production that is `railThreadsUnread(db)`, which runs `ListSubscribedThreads` and reports whether any row is unread. A workspace lights if either a visibly-unread channel or an unread subscribed thread exists. A workspace the router doesn't know yet (still connecting) keeps #207's "light it" default: the self ID is unknown, so the self-authored suppression can't fire, and any unread thread lights it. Threads are not subject to channel mute. Reusing the query rather than writing an `EXISTS` twin is deliberate, per your note on #207 about a third parallel unread rule. Cost was measured first: five workspaces with at most ten active subscriptions each answer in well under a millisecond total on my cache, and the reader only runs on read-state events.
- **The UI recomputes the rail when thread read state changes.** Three sites in `internal/ui`, chosen by one rule, *the DB has already changed when this message arrives*: `applyThreadMarkListState` (both `ThreadMarkedRemoteMsg` and `ThreadMarkedLocalMsg`; the cursor is written before either is sent), the `ThreadsListLoadedMsg` arm (active workspace so the rail moves with the badge; inactive because that message is the tail of a change dispatched before the user switched away, and the switch itself recomputes nothing), and `reduceNewMessage`'s thread-reply arm, mirroring its channel arms including the "on screen and focused, the staged mark will clear it" exception. The optimistic badge sites (`MarkSelectedRead` on open, the `ThreadRepliesLoadedMsg` recompute, `applyThreadMarkUnread`) deliberately do not fan out: nothing in the DB has changed at those moments, so the rail would only re-read the answer it already shows; the mark they issue, or its `thread_marked` echo, refreshes it when it lands.
- **Inactive-workspace thread events reach the rail.** `OnThreadMarked` and `OnThreadSubscriptionChanged` already persisted for inactive workspaces but returned before sending anything. Each inactive branch now sends `ReadStateChangedMsg`, the same choice `muteRefreshMsg` makes. To make that assertable, `rtmEventHandler.program` is narrowed from `*tea.Program` to the `teaSender` interface `reconnect_sync.go` already uses (only `Send` was ever called through it, and `bootstrapPresenceAndDND`'s parameter goes the same way).
- **The subscription reconcile reaches the rail.** Boot, reconnect and wake all run `ensureWorkspaceThreadSubs`, which rewrites `last_read`/`latest_reply` for every workspace and signals only `ThreadsListDirtyMsg`; that arm returned at once for an inactive workspace. So the most common way thread state changes, read in another client while slk was closed, never reached the dot: `WorkspaceReadyMsg`'s refresh runs before the staggered sweep and sees the old rows. The inactive arm now recomputes the rail and still schedules no fetch. Both reviewers found this independently.
- The `railUnreadWorkspaces` doc comment is trimmed to the rule and its edges (and says why an empty self ID cannot suppress: cached replies always carry a user or bot ID), and one more test names the decision that #207's channel-membership rule does not apply to threads.

`$SLK_UNREAD` and the title's `(N)` still count channels in the active workspace only, as `wiki/Configuration.md` documents. The dot, `+N` and `$SLK_OTHER_UNREAD` are "workspaces with something unread", and threads fit that; the wiki line for `$SLK_OTHER_UNREAD` says so now. If you'd rather threads count in `(N)` too, that's a one-line follow-up.

## Why not fix it elsewhere

- **The boot-time `counts.threads.has_unreads` flag** (`wctx.ThreadsHasUnreads`)? It's stale from the first `thread_marked` onward; the code already retired it in favour of per-thread `last_read` (see the comment on the `ListFetch` closure in `main.go`). Using it here would make the dot drift from the badge.
- **A `has_unread_threads` column on `workspaces`?** A migration for a derived value.
- **Read the in-memory badge?** It exists only for the active workspace.
- **A cheaper `EXISTS` query?** Not needed at the measured cost, and it would be a second predicate to keep in step with the first.

## Tests

Plain `testing.T`, white-box. Nothing shared beyond the file-local helpers listed.

- `TestRailUnreadWorkspaces_Threads` (`cmd/slk/rail_unread_test.go`): through the real cache, seeded with `ReconcileThreadSubscriptions` (the only writer of `latest_reply`) + `UpsertMessage` (helper `seedSubscribedThread`). Unread by the getView watermark → lit; unread by a cached reply newer than the watermark → lit (this is the field shape: `latest_reply == last_read`, newer reply in the cache, which a plain `latest_reply > last_read` query misses); newest cached reply self-authored → dark; fully read → dark; thread in a channel absent from `wctx.Channels` → lit (the badge does not filter on the list either, so #207's membership rule stays a channel-row rule); muted unread channel plus unread thread → lit. The thread half never looks at channel type, so DM threads, which the field case was, light the same way.
- `TestRailUnreadWorkspaces_ThreadsUnknownWorkspace`: router doesn't know the workspace, unread thread → lit.
- `TestRailUnreadWorkspaces_RailAndTitleAgree` extended with a thread-only workspace: the dots and `OtherUnreadCount` are still the same set.
- `TestThreadReadState_RefreshesRail` (`internal/ui/reducer_threads_unread_refresh_test.go`): through `App.Update`, each of `ThreadMarkedRemoteMsg`, `ThreadMarkedLocalMsg`, `ThreadsListLoadedMsg` (active) and a thread-reply `NewMessageMsg` recomputes the rail and title.
- `TestThreadsListLoaded_InactiveWorkspace_RefreshesRail` and `TestThreadsListDirty_InactiveWorkspace_RefreshesRail`: the two arms that used to return before doing anything now recompute the rail; the first still leaves the badge alone, the second still schedules no fetch.
- `TestThreadReply_OnScreenAndFocused_DoesNotRefreshRail`: the one exception.
- `TestOnThreadMarked_InactiveWorkspaceRefreshesRail` and `TestOnThreadSubscriptionChanged_InactiveWorkspaceRefreshesRail` (`cmd/slk/event_handler_test.go`): exactly one `ReadStateChangedMsg` is dispatched, and none of the active-only messages.

Each commit's tests were confirmed red with its change reverted. To repeat: for commit 1, remove the `teamIDs` loop in `railUnreadWorkspaces` (the signature has to stay or the tests don't compile) and `TestRailUnreadWorkspaces_Threads`, `_ThreadsUnknownWorkspace` and `_RailAndTitleAgree` go red; for commit 2, drop its three `notifyReadStateChanged` calls and `TestThreadReadState_RefreshesRail` plus `TestThreadsListLoaded_InactiveWorkspace_RefreshesRail` go red; for commit 3, drop the two sends (keep `teaSender`) and both `*_InactiveWorkspaceRefreshesRail` handler tests go red; for the reconcile commit, drop its one call and `TestThreadsListDirty_InactiveWorkspace_RefreshesRail` goes red. `go build`, `go vet`, `go test ./... -race`, `gofmt -l` and `golangci-lint run` (v2.13.1) are all clean.

**Field check (macOS, five live workspaces):** with the patched build sitting on another workspace, the workspace with two unread DM thread replies and no unread channel showed a lit dot on launch (the boot path: cached rows, then the reconcile). Reading both threads in the official client turned the dot off without touching slk (the `OnThreadMarked` inactive-branch send). A new reply in one of them lit it again (the `OnMessage` inactive path, which already sent `ReadStateChangedMsg`). That is the thread rule and both inactive-workspace refreshes end to end. The two live threads were unread by the getView-watermark half at that point (their `latest_reply` had been reconciled forward), so the live run exercises that shape; the cached-reply shape is the second case of `TestRailUnreadWorkspaces_Threads`.

Things that usually bite, checked: DM threads light the same as channel threads (the thread half never reads channel type; the field case is a DM thread). #207's unlisted-channel rule does not hide a thread (tested). The nil-`wctx` boot default stays conservative (show, don't hide). There is no third unread predicate: the rail calls `ListSubscribedThreads`. Every reply in the active workspace now recomputes the rail, subscribed thread or not, at the measured cost above. New reads on the UI goroutine are `wctx.UserID` (written once at construction, before `router.Add`) and the configured team list (built once before the reader is installed); no new reader of `wctx.Channels`.

## Noted but not changed

- **Cost at scale.** The reader runs `ListSubscribedThreads` once per configured workspace on every read-state event, synchronously on the UI goroutine, and `notifyReadStateChanged` calls the reader twice (`RefreshUnreads`, then `OtherUnreadCount`; that double call predates this PR). On my cache that is under a millisecond in total. Codex built a synthetic fixture of 1,000 active subscriptions and 100,000 cached messages in one workspace and measured ~365 ms per query there. A user with that many subscriptions already pays that query on every threads-list refresh of the active workspace today; what this PR adds is the multiplication by workspaces and by the two reader calls, on read-state events, which are more frequent. If a cache of that shape turns up, the cheap half of the follow-up is to have `OtherUnreadCount` count the rail's own `HasUnread` flags that `RefreshUnreads` just set instead of calling the reader again, which halves the cost, makes "rail and title agree" structural, and removes the pre-existing quirk where `+N` counts a workspace that is in the cache but not in the rail. The second half, only if a real cache asks for it, is an `EXISTS ... LIMIT 1` form of the same SQL pinned to `ListSubscribedThreads` by a lockstep test. Neither is done here: the first is outside this diff, the second is a second predicate, and nothing measured asks for either yet.
- `reduceWorkspaceSwitched` does not call `notifyReadStateChanged`, so a workspace switch itself never recomputes the rail or title; they catch up on the next read-state event. Pre-existing, and the reason the inactive `ThreadsListLoadedMsg` and `ThreadsListDirtyMsg` arms refresh rather than return.
- `wctx.Channels` is still read on the UI goroutine without a lock (#208); this PR adds no new reader of it.
- `ListFetch` computes summaries with the *active* workspace's user ID even when asked for another team. Harmless today (the reducer drops the other team's list), and the rail reader passes each workspace its own ID.

## AI assistance

Written with Claude (Fable 5.1, extra-high thinking effort) driving the investigation, code and tests from a brief I wrote after reproducing the bug on live workspaces. The cause was confirmed against a live `client.counts` response and the cache's actual rows rather than inferred from the code, and Codex reviewed the diff. I've read the diff and will defend it in review.


gammons:
Same class of bug as #207, one signal over, and the same correct fix shape. Merging.

A thread reply never sets the channel's `has_unread`, so the channel-row pass can't see it — the only way a thread reaches the rail is by checking the subscribed-thread read cursor, which is exactly what the sidebar's Threads badge counts. You've extended the single-predicate approach from #207 rather than inventing a parallel rule, and you're right that channel mute and `wctx.Channels` membership don't apply to thread unreads (they don't apply to the sidebar badge either).

I verified the tests have teeth: disabling the thread arm in `railUnreadWorkspaces` fails both `TestRailUnreadWorkspaces_Threads` and `TestRailUnreadWorkspaces_ThreadsUnknownWorkspace`.

And I can reproduce why you hit it — the #207 mute fix landing is what made a thread-only-unread workspace go dark. Good catch.

One note: GitHub shows this as CONFLICTING, but `git merge-tree` against current `origin/main` merges cleanly — that's stale mergeability from when the PR was created against an older main. It'll clear on the next push or on merge.

Merging.

gammons:
@landonforshage I merged `main` into your branch to clear the conflict — it was a real one this time, not the stale GitHub computation.

The conflict was in `cmd/slk/main.go`: #219's refactor wrapped the `unreadWorkspaces` closure in `core.NewUnreadService` and replaced `App.SetWorkspaceUnreadReader` with `App.SetUnreadService`, landing right on top of where you'd changed the call to take `railTeamIDs` and `railThreadsUnread(db)`.

The resolution keeps your 4-arg `railUnreadWorkspaces` call inside the new `SetUnreadService` wiring. `SetUnreadService` still routes the closure to `workspaceRail.SetUnreadReader` (`internal/ui/app.go:2458`), so the rail/title/`$SLK_OTHER_UNREAD` behavior is unchanged.

Two things needed touching beyond the conflict hunk:

- `internal/ui/reducer_threads_unread_refresh_test.go` referenced the old `SetWorkspaceUnreadReader`. It's a white-box test in package `ui`, so it now installs the rail reader directly via `app.workspaceRail.SetUnreadReader` — same effect, no API change.
- Updated the stale `rail_unread.go` comment to name `SetUnreadService`.

Verified: build, vet, gofmt clean, and `go test ./... -race` is 57 packages, 0 fails — including both thread rail tests.

CI is green. This is ready to merge.

PR #219, opened 2026-09-15
feat: show other people's status, DND and huddles

### LINKED ISSUE #211: Show other people's custom status, DND and huddle state
Image

## Problem

slk currently shows only online/away presence for people in direct messages. It does not show whether somebody is in a meeting, on vacation, in Do Not Disturb or in a huddle, so users have to open Slack's official client to recover context that is normally visible beside a person's name.

## Proposed behavior

- Show a person's custom-status emoji beside their name on DM rows, message authors, thread authors and channel-finder results.
- While they are in a huddle, show a headphones indicator in that same slot.
- Replace the DM presence marker with a DND indicator while that person is in DND. Match Slack by not adding DND beside message authors.
- Show the full custom-status text and DND end time in an open DM's header.
- Remove status and DND indicators when they expire.
- Keep these values isolated per workspace and preserve them across sidebar rebuilds and workspace switches.

This is display-only. Setting somebody's status, joining or starting huddles, and adding DND beside message authors are out of scope.

## How Slack reports the state

Observed from debug logs of live browser-protocol sessions:

- Other users' profile changes arrive as `user_invalidated`; the event carries only the user ID, so the client must refetch the profile.
- DND changes arrive as `dnd_invalidated`, also carrying only an ID; current state comes from `dnd.teamInfo`.
- `huddle_state` is `in_a_huddle` during a huddle and `default_unset` otherwise.
- A huddle join produced `user_invalidated`, but one of two observed huddle endings produced no event. A bounded periodic recheck is therefore needed while anybody is believed to be in a huddle.
- In a 12-hour sample, 328 `user_invalidated` events covered 65 users. Coalescing short bursts avoids issuing one request per event.

## Proposed implementation

1. Decode `user_change`, `user_status_changed`, `user_invalidated`, `dnd_invalidated` and `dnd_updated_user` without routing another user's DND into the signed-in user's status bar.
2. Coalesce invalidations for a short window, refetch known users through the existing batched users/info path, and split `dnd.teamInfo` calls at its documented 50-user limit.
3. Refetch DM peers' DND on connect and workspace switch because the socket does not replay missed invalidations.
4. Recheck only users last observed in a huddle, once per minute, until a refetch reports that the huddle ended.
5. Persist custom status and huddle fields in the existing SQLite user cache. Reset cached user versions once when the new columns appear so conditional revalidation backfills existing rows. Slack remains authoritative.
6. Keep glyph selection, expiry and DM-header formatting in one UI package shared by the sidebar, finder, messages and thread panes.
7. Deliver first-sight status/huddle data when an unknown message author is resolved, and gate all live updates by workspace.

The first start after the cache migration will refetch every cached user once. Steady-state request volume is one users/info batch per invalidation burst, one `dnd.teamInfo` call per 50 relevant users at connect/switch or per invalidation burst, and one users/info batch per minute only while at least one user is believed to be in a huddle.

## Implementation status

A working local implementation already exists and exercises the approach above end to end. It is intentionally not opened as a PR yet: the change is about 3.1k lines across 45 files, so this issue is the design checkpoint before asking for review. I am happy to revise or split the implementation based on the discussion here.

The implementation and its regression tests currently pass the repository's release build, vet, full test suite, race suite, formatting and lint checks. The separate kitty upload fix is kept as its own commit so the feature branch can rebase to one feature commit after that fix lands.

## Relationship to existing work

This overlaps issue #78 only at the invalidation/refetch mechanism. It should not claim to fix #78 without proving that rename and avatar changes propagate all the way through the cache and UI. #78 also covers bot identity and a configurable TTL, which are outside this proposal.

Open PRs #167 and #168 edit some of the same composition, resolver, workspace-reducer and sidebar files. The implementation should rebase after whichever of those lands; this is conflict risk, not duplicate functionality.

A separate kitty image-upload fix should land first. Peer-status refetches can preload avatars mid-session, increasing concurrency on the kitty graphics side channel.

## Acceptance criteria

- Custom status and huddle indicators render on DM rows, finder rows, main-message authors and thread authors.
- DND replaces presence on DM/finder rows and appears in the open-DM header, but not beside message authors.
- Live changes update every active surface without leaking across workspaces.
- Expired state disappears without requiring another Slack event.
- A missed huddle-ending event is corrected by the bounded recheck.
- Cache migration backfills existing users once and remains idempotent.
- First-sight resolution carries status/huddle state into both cache and UI.
- `dnd.teamInfo` never receives more than 50 users.
- Regression tests fail when event routing, cache backfill, rebuild preservation, surface rendering, huddle recheck, first-sight delivery or DND batching is removed.

rfist:
Closes #211.

## Problem

The Slack client shows at a glance that a teammate is in a meeting, on vacation, in Do Not Disturb or in a huddle. slk showed only the online/away dot for DM peers, so none of that was visible.

This PR shows custom status and huddle indicators on DM rows, next to message authors (main pane and threads), in the channel finder, and in an open DM's header. DND replaces the presence marker on DM rows and finder results and also appears in the DM header; Slack does not show it beside message authors.

**It reverses one existing contract, so that first.** `dnd_updated_user` was dispatched to the self-DND handler together with `dnd_updated`, and `TestDispatchWebSocketDNDUpdatedUserEvent_NoDND` asserted that. The event is about another user, so routing it there would let a teammate's DND change slk's own status-bar segment. It now goes to a new `OnUserDNDChange`, and one that names no user is dropped (`internal/slack/events.go:452`). The old test became `TestDispatchWebSocketDNDUpdatedEvent_NoDNDActive` for `dnd_updated`. The event was never seen on the socket in the captures below, so this was latent rather than something users hit.

## How Slack reports other users' state

Measured from debug logs of real sessions (counts only):

- The browser-protocol socket does not push other users' `user_change`. Over 12 hours of one workspace, every `user_change` and `user_status_changed` came from a single user, the signed-in one as far as the log can tell.
- Other users' profile changes arrive as `user_invalidated` (328 events across 65 users in those 12 hours) and DND changes as `dnd_invalidated`. Both carry only `{type, event_ts, user:{id}}`, so the client has to refetch.
- `huddle_state` is `in_a_huddle` during a huddle and `default_unset` otherwise, with `huddle_state_expiration_ts` about 22 minutes after joining. `user_invalidated` arrived for every huddle join observed, but for one of two huddle endings no event arrived at all.

## Changes

**Events.** `user_change` / `user_status_changed` decode the status and huddle fields into a `UserStatus` (`internal/slack/events.go:463`). `user_invalidated` / `dnd_invalidated` hand the ID to the handler (`internal/slack/events.go:471`). `SlackAPI` gains only `GetDNDTeamInfoContext`.

**Refetch.** `peerStatusRefresher` (`cmd/slk/peer_status.go:56`) coalesces invalidations for 2 s into one call to the existing `userResolver.ResolveNow` (edge users/info, at most 80 IDs per request, `internal/slack/edge/cache.go:45`), which already writes the cache and emits the status for every user it returns. Users it does not return, including all of them when edge is degraded or the call fails, are refetched one at a time over users.info (`cmd/slk/peer_status.go:177`). Both paths end in exactly one status message per user; the users.info side goes through `applyProfileStatus` (`cmd/slk/peer_status.go:201`). It refetches only users already in the cache and never the signed-in user, whose own events carry the state. Users last seen in a huddle are refetched every minute until a result says the huddle ended (`cmd/slk/peer_status.go:37`, `cmd/slk/peer_status.go:255`).

The first-sight resolver persists and emits the status and huddle fields when an unknown message author becomes visible (`cmd/slk/main.go:464`, `cmd/slk/main.go:654`), and so does the per-user fallback of the unresolved-DM sweep (`cmd/slk/main.go:2995`, `cmd/slk/main.go:3024`).

**DND.** `dnd.teamInfo` requests are split at Slack's documented 50-user limit; a failed batch leaves only its own users unknown (`cmd/slk/peer_status.go:282`). Within a burst of invalidations, DND is refreshed before the per-user profile fallback, with a 2 s bound, so a slow users.info call cannot hold it up (`cmd/slk/peer_status.go:153`). DM peers' DND is fetched on connect (`cmd/slk/main.go:4116`), because the socket replays nothing, and after a workspace switch. The switch itself does not wait for it: `WorkspaceSwitchedMsg.RefreshPeerDND` (`internal/ui/msgs.go:293`, built at `cmd/slk/main.go:2075`) is appended to the switch reducer's batch after `ResetPresence` has cleared the previous workspace's live state (`internal/ui/reducer_workspace.go:390`), and its results carry the team ID, so a result that arrives while another workspace is active is dropped.

Request volume: one `ResolveNow` per burst of invalidations (plus one users.info call per user it misses, which is every user in the burst while edge is degraded), one `dnd.teamInfo` per 50 DM peers at connect and switch, one call per 50 users in a `dnd_invalidated` burst, and one refetch per minute only while someone is in a huddle.

**Cache.** `users` gains `status_emoji`, `status_text`, `status_expiration`, `huddle_state` and `huddle_expiration`. Conditional revalidation only returns users whose version moved, so users cached before the columns existed would never be refetched to fill them: the migration zeroes `users.version` once for the status columns and once for the huddle columns, whenever any column of that group is missing, and does so before adding them, so an interrupted migration resets again on the next start (`internal/cache/db.go:265`, `internal/cache/db.go:290`). **Blast radius:** the first start after upgrading refetches every cached user once.

`UpsertUser` writes the new columns on insert but not on conflict (`internal/cache/users.go:40`), because several callers upsert placeholder records that carry no profile and would blank a real status. `EdgeUserUpdate` always writes them, empty included: unlike `image_original`, users/info carries these keys empty when nothing is set.

**UI.** A new `internal/ui/peerstatus` package decides expiry, glyph resolution and the DM header summary once for all four surfaces. The status emoji resolves through `emoji.CodeMap` after `StripSkinTone`; a custom workspace emoji or a text-only status shows 💬, and 🎧 takes the emoji's place during a huddle.

- DM rows start from the cached status, both at startup and when a DM opens mid-session (`seedDMFromCache`, `cmd/slk/channelitem.go:15`, called at `cmd/slk/main.go:2743` and `cmd/slk/main.go:4951`).
- Sidebar: `statusByUser` follows `presenceByUser`'s contract of surviving `SetItems` rebuilds (`internal/ui/sidebar/model.go:979`). `⊘` replaces the presence dot during DND, and the emoji's width is charged to the name budget so long names truncate instead of wrapping.
- Channel finder: statuses live in a map keyed by channel ID (`internal/ui/channelfinder/model.go:68`) rather than on `channelfinder.Item`, fed from the sidebar when a workspace becomes active (`internal/ui/presence.go:305`) and on every change, and pruned to the current rows (`internal/ui/channelfinder/model.go:88`). The glyph is charged to the name's budget the same way as in the sidebar (`internal/ui/channelfinder/model.go:640`).
- Message and thread panes: the emoji follows the author name (`internal/ui/messages/model.go:1466`), and the render cache is only invalidated when that user authored something in the pane.
- The presence reducer handles the new messages (`internal/ui/presence.go:195`), updates the header of every window showing the DM, not only the focused one (`internal/ui/presence.go:329`), and runs a once-a-minute expiry tick only while some status, DND or huddle has a deadline (`internal/ui/presence.go:349`).

**Docs.** `wiki/Features.md` and `docs/STATUS.md` describe the indicators. The "Not planned" line in `wiki/Tradeoffs-and-Non-Goals.md` now says joining or starting huddles, since slk shows who is in one. `AGENTS.md` lists `slack.DNDStateFromStatus` and `ui/peerstatus` in the shared-helper table.

## Ordering with #210, #212 and #162, and the refactor plan

**#210 should merge first.** This branch does not include it and does not need it to build, but refetches preload avatars mid-session, which made that kitty interleaving bug frequent enough that avatars showed other images.

**#212 conflicts on one import line.** A trial merge against #212's head auto-merges `main.go`, `channelfinder/model.go`, `messages/model.go`, `sidebar/model.go` and `AGENTS.md`. The only conflict is in `internal/ui/msgs.go`, where this PR adds the `tea` import directly above the `internal/cache` import that #212 replaces with `internal/core`; the resolution keeps both. This PR does not touch `channelfinder.Item`, `callbacks.go` or `app.go`, so #212's alias and ports need nothing from it. Whichever lands second rebases.

**#162** adds a DM opened mid-session to the channel finder. Until it lands, such a DM has a sidebar row (with its cached status) but no finder row; after it lands, the finder row's status appears with the next status change.

**Against `2026-09-06-architecture-refactor.md`:**

- Phase 1: the refetch logic is in its own file, `cmd/slk/peer_status.go`; what this PR adds to `main.go` is wiring.
- Phase 2: the refresher's `send` callback uses the same `if p != nil { p.Send(msg) }` capture as `newUserResolver` and `membership.New` beside it, so step 4 replaces all three the same way. The switcher still hands `wctx.Channels` to the UI while WebSocket handlers write it, exactly as on `main`; this PR adds no lock around that shared state and leaves it to steps 1–3.
- Phase 3: `SetUserStatuses`, `PatchUserStatus`, `ExpireStatuses` and `hasAuthor` exist on both `messages.Model` and `thread.Model`. This adds to finding F3; the rendered suffix itself is shared (`messages.AuthorStatusSuffix`), and the one real difference is that the thread pane also bumps `userNamesV` because its chrome cache renders the parent's header. These should collapse into the shared pane in Phase 3.

It also touches the same resolver, workspace reducer and sidebar files as #167 and #168; whichever lands first, the other rebases.

## Not covered

- Setting your own custom status.
- DND next to message authors (Slack does not show it there either).
- Joining or starting huddles.
- Right after a workspace switch, DM rows show no DND until `dnd.teamInfo` returns (bounded by 2 s). A result from an earlier visit that is still in flight when the user comes back to the same workspace passes the team-ID check, the same gating the threads list uses.
- slk's API HTTP client sets no request timeout, which predates this PR. While edge is degraded the per-user fallback makes its users.info calls one at a time, so a hung call delays the rest of that burst's status updates (not its DND, which is fetched first).
- Whether Slack extends `huddle_state_expiration_ts` during a huddle longer than about 22 minutes is not verified. The per-minute recheck picks up a renewed expiry; if Slack does not renew it, the 🎧 disappears after about 22 minutes.
- The wiring in `run()` and `connectWorkspace` has no test. Replacing the refresher's construction with a discard leaves the suite green; as the refactor plan notes, nothing covers those two functions yet. The tests below drive the refresher, resolvers, event handler, cache and reducers directly.

## Tests

Added by this diff:

- `internal/slack/events_test.go`: `TestDispatchWebSocketDNDUpdatedUserEvent_RoutesToPeerNotSelf`, `TestDispatchWebSocketDNDUpdatedUserEvent_WithoutUserIsDropped`, `TestDispatchWebSocketUserChangeEvent_DeliversStatus`, `TestDispatchWebSocketInvalidationEvents`, `TestDNDStateFromStatus_MatchesComputeDNDState`
- `internal/cache/user_status_test.go` and `internal/cache/user_huddle_test.go`: status and huddle writes on every path, insert-only upsert, `ListUserStatuses`, and both migrations' one-time version reset
- `internal/cache/db_test.go`: `TestMigration_StatusColumnGroupResetsVersionEvenIfOneColumnAlreadyExisted`, `TestMigration_HuddleColumnGroupResetsVersionEvenIfOneColumnAlreadyExisted`, `TestMigration_StatusColumnGroupLeavesVersionAloneWhenAlreadyComplete`
- `cmd/slk/peer_status_test.go`: coalescing, self and unknown-user skipping, 50-user DND batching and a failing later batch, the users.info fallback when edge returns nothing or only some users, one status message per user, DND fetched before the profile fallback, huddle recheck until the huddle ends, seeding, cache seeding helpers
- `cmd/slk/user_resolver_test.go`: first-sight users.info and edge resolution persist and emit status and huddle state; `TestResolveDMNames_FallbackPersistsStatusAndEmitsMessage` drives the unresolved-DM sweep through its per-user fallback for a user not yet cached
- `cmd/slk/event_handler_test.go`: `TestOnConversationOpened_SeedsDMStatusFromCache`
- `internal/ui/peerstatus/status_test.go` and `huddle_test.go`: glyph resolution, expiry, summary, huddle precedence
- `internal/ui/sidebar/status_test.go`, `status_view_test.go`, `huddle_test.go`: survive-rebuild contract, expiry, and the rendered row
- `internal/ui/messages/model_test.go`, `internal/ui/thread/model_test.go` and `internal/ui/channelfinder/model_test.go`: status and huddle rendering beside authors and in finder rows, including DND in the finder, a long name keeping its glyph, and finder statuses surviving `SetItems` but not their rows
- `internal/ui/app_peer_status_test.go` and `app_peer_huddle_test.go`: reducer fanout, tick scheduling, inactive-workspace gating, channel-selection wiring, live DM-header updates, and header updates and expiry in an unfocused split window
- `internal/ui/reducer_workspace_peerdnd_test.go`: the switch reducer runs `RefreshPeerDND` after applying the switch, and its result reaches the DM row

Changed: `TestDispatchWebSocketDNDUpdatedUserEvent_NoDND` (see the top), and the `TestConversationsView_DecodesUsers` fixture, whose `huddle_state` values were `WRONG-…` placeholders for an unmodelled key and are now distinct modelled values.

Revert checks, each run on its own:

- routing `dnd_updated_user` back to the self handler fails `TestDispatchWebSocketDNDUpdatedUserEvent_RoutesToPeerNotSelf`
- dropping huddle tracking from the refetch fails `TestPeerStatus_RechecksHuddlersUntilTheHuddleEnds`
- removing the per-user users.info fallback fails `TestPeerStatus_FlushFallsBackWhenEdgeReturnsNil` and `TestPeerStatus_FlushFallsBackOnlyForIDsEdgeMissed`; emitting again for users edge returned fails `TestPeerStatus_CoalescesKnownPeersIntoOneRefetch` ("got 4 status msgs; want 2")
- refreshing DND after the profile fallback fails `TestPeerStatus_FlushRefreshesDNDBeforeProfileFallback`
- dropping status persistence from the unresolved-DM fallback fails `TestResolveDMNames_FallbackPersistsStatusAndEmitsMessage`
- not seeding a DM opened mid-session fails `TestOnConversationOpened_SeedsDMStatusFromCache`
- writing status in `UpsertUser`'s conflict clause fails `TestUpsertUser_WritesStatusOnInsertButNotOnConflict`
- not reapplying remembered status on rebuild fails `TestStatusByUser_AppliedLateAndSurvivesRebuild`
- skipping the huddle migration's version reset fails `TestHuddleMigration_ResetsUserVersionsOnDatabasesWithStatusColumns`; checking only the first column of each group fails both `…EvenIfOneColumnAlreadyExisted` tests
- removing the 50-user `dnd.teamInfo` split fails `TestPeerStatus_FetchDNDBatchesAtSlackLimit`; returning nothing when a later batch fails fails `TestPeerStatus_FetchDNDKeepsEarlierBatchesWhenALaterOneFails`
- dropping first-sight status persistence or UI delivery fails `TestUserResolver_FirstSightPerUserDeliversPeerStatus` and `TestUserResolver_FirstSightEdgeDeliversPeerStatus`
- removing the author suffix fails `TestMessagePaneViewUpdatesAuthorStatus` and `TestThreadViewUpdatesAuthorStatus`
- removing the finder suffix fails `TestViewShowsPeerStatusAndDND`; truncating the whole finder row instead of the name fails `TestRenderBoxTruncatesNameNotGlyph`; not pruning finder statuses fails `TestSetItemsDropsStatusesForRemovedRows`
- restoring the blank topic on channel selection fails `TestChannelSelectedAndLivePeerChangesUpdateDMHeader`; updating only the focused window's header fails `TestUnfocusedSplitWindowGetsLivePeerStatusHeader` and `TestUnfocusedSplitWindowStatusExpiryUpdatesHeader`
- not appending `RefreshPeerDND` in the switch reducer fails `TestWorkspaceSwitched_RefreshesPeerDNDAfterSwitchApplies`

## Test plan

- `gofmt -l $(git ls-files '*.go')`
- `go vet ./...`
- `go test -count=1 ./...`
- `go test -race ./...`
- `git diff --check main...HEAD`
- Revert checks above
- Trial merge against #212's head (`git merge-tree`), result above
- Manual, against a live workspace next to the official client: custom status emoji on DM rows and message authors, huddle joins and endings (including an ending that sent no event), and statuses seeded from the cache after a restart
- One commit on current `main` (`356d29e`)


gammons:
Reviewed closely — this is a large PR, and it holds up. Merging.

**The DND contract reversal is the part that needed scrutiny, and it's handled correctly.** `dnd_updated_user` is about another user; routing it into `OnDNDChange` would let a teammate's DND flip slk's own status-bar segment. You've split it into `OnUserDNDChange`, the test was *inverted* rather than deleted (`TestDispatchWebSocketDNDUpdatedUserEvent_RoutesToPeerNotSelf` now asserts the peer event reaches the new handler and never `OnDNDChange`), and there's a second test for the no-user drop case. Nothing else depended on the old routing.

**On API cost — this was my main worry at 3753 lines, and it's disciplined:**

- Boot: **zero** new status requests. Statuses read from SQLite and ride along on the already-fetched boot payloads.
- Connect: **one** `dnd.teamInfo`, batched over self + DM peers at Slack's 50-user limit, with a timeout, async.
- Runtime invalidations coalesce into a 2s window and flush as one batched `users/info` + `dnd.teamInfo` — not a per-user fan-out. The `known` gate drops users not already in the cache, which is the exact opposite of the `users.info` boot-storm we fought in the membership resolver.

**On freshness:** event-driven off `user_change`/`user_status_changed`/`user_invalidated`/`dnd_invalidated`, with a one-per-minute huddle recheck only while the huddler set is non-empty (since Slack doesn't reliably signal huddle end). No blind polling.

**On the non-goal:** huddles-as-in-hosting are the non-goal; this is displaying huddle state, which is the same availability signal as custom status and core to interpreting a teammate. Amending the wiki line to "slk only shows who is in one" is the right call.

No keybinding changes, no golden files touched, composes correctly with the #203 background discipline and the colored-usernames work. I spot-checked two tests by reverting the production change; both fail as designed.

Two polish notes, not blockers: `peerstatus.Status` is `==`-compared in a few `Patch*` paths — fine today (all `time.Time`s come from `time.Unix`, no monotonic component) but a one-line comment would future-proof it. And a status that changed while slk was off appears at next workspace switch rather than at boot-day revalidation — acceptable.

GitHub shows CONFLICTING but `git merge-tree` against current `main` merges cleanly — stale mergeability from the PR's original base. It'll clear on push or merge.

Merging.

PR #221, opened 2026-09-16
fix(slk): close the activeTeamID and cfg.Workspaces connect races

laraibg786:
review of #212 flagged following concurrency defects in `main.go`'s closure-capture style as adjacent, pre-existing bugs.

1. `activeTeamID` — a plain `string` written from the UI goroutine and from N connect goroutines, read from every WebSocket goroutine via `isActive`.
2. `cfg` mutated in place by `SetThemeSaver`/`SetWidthSaver` on the UI goroutine while a copy sharing `cfg.Workspaces` lives in every `rtmEventHandler` and it turned out to be broader it's not just the theme/width savers racing `rtmEventHandler.cfg` — `buildChannelItem`'s section lookups and `ResolveTheme`/`ResolveWidth` inside `connectWorkspace` itself also read the same shared `cfg.Workspaces` map, from every connect goroutine, concurrently with the Update goroutine's writes.

## Commit Description

Final state, upfront: `activeTeamID` no longer exists in any form — every site reads `router.Active()` directly. `cfg.Workspaces` is now owned exclusively by `workspaceConfigStore`, mutex-guarded, with callers reading an independently-cloned snapshot. `activeTeam{id string}` below is transient scaffolding that existed only long enough to prove the race, then was deleted — it is not part of the final design.

`activeTeamID`/`cfg.Workspaces` lived entirely inside closures local to `run()`. There was no way to demonstrate either race against a test without extracting the racy logic into something callable first. So this is structured as a real red→green:

1. **`refactor(slk): extract activeTeamID and cfg.Workspaces into named types`** — pulls the plain `activeTeamID string` into `activeTeam{ id string }` (scaffolding only, deleted in commit 3) and the inline `cfg.Workspaces` read-modify-write into `workspaceConfigStore` (`SaveTheme`/`SaveSidebarWidth`/`Snapshot`, which does persist). Purely mechanical: neither type adds synchronization yet, and `Snapshot` still returns a value copy of `cfg` that aliases the same `Workspaces` map — the exact shape of both races, just now expressed as something a test can call.
2. **`test(slk): add failing -race tests for activeTeam and workspaceConfigStore`** — `TestActiveTeam_ConcurrentSetAndGet` and `TestWorkspaceConfigStore_ConcurrentSaveAndSnapshot`, both reproducing the exact pattern boot produces. **`go test -race ./cmd/slk/...` fails on this commit** — confirmed before committing it, not asserted after the fact.
3. **`fix(slk): close the activeTeam and workspaceConfigStore races`** — three fixes:
   - `activeTeam` is **deleted**, not synchronized. It duplicated `router.Active().TeamID`, already `atomic.Pointer`-backed and therefore race-safe on its own. Every read site now calls `router.Active()` directly; every write site already called `router.Set(wctx)` immediately alongside the old `active.Set(...)` line, so deleting the latter changes nothing behaviorally. `TestActiveTeam_ConcurrentSetAndGet` is deleted with it and replaced by `TestWorkspaceRouter_ConcurrentSetAndActive` (pinning that `Set`/`Active` — which the fix now relies on exclusively — actually is safe; it had no concurrency coverage of its own before this).
   - `workspaceConfigStore` gets a **real fix**: a `sync.RWMutex` around `SaveTheme`/`SaveSidebarWidth`'s read-modify-write, and `Snapshot` now returns an independent copy of `Workspaces` instead of aliasing the map those methods mutate. `TestWorkspaceConfigStore_ConcurrentSaveAndSnapshot` — red in commit 2 — passes unchanged in commit 3; no test edits needed to go from red to green.
   - The workspace-switcher closure (`core.NewWorkspaceService`'s callback, runs on a `tea.Cmd` goroutine) still read the **outer** `cfg` directly for `Theme`/`SidebarWidth` instead of `workspacesStore` — same shape of hole, caught in review. When `cfg.Workspaces` was non-nil at boot this raced the store's mutex-guarded writes with no mutex on the read at all; when it started nil, the store's private map meant the switcher's saved theme/width could silently never be seen this session (self-heals on restart). Routed both reads through `workspacesStore.Snapshot()`, same pattern the connect path already used. No new test: `TestWorkspaceConfigStore_ConcurrentSaveAndSnapshot` already pins `Snapshot()` racing `SaveTheme`, the exact contract this depends on.

## Verification

- `go build ./...`, `go vet ./...`: clean on **all three commits** individually (checked out and rebuilt each one, not just the branch tip).
- `go test -race -count=1 ./...`: all 56 packages pass on commits 1 and 3; commit 2 fails exactly as intended (`TestActiveTeam_ConcurrentSetAndGet` and `TestWorkspaceConfigStore_ConcurrentSaveAndSnapshot`), confirmed via explicit `-run` before committing it.
- No behavior change intended: `router.Set`/`Active` were already called everywhere `activeTeamID`/`activeTeam` was set, in the same order; `cfg.Workspaces`' read-modify-write shape is unchanged, just now mutex-guarded and read from an independent snapshot instead of the shared map.
- `golangci-lint run ./...` (v2.13.1, matching CI): 0 issues on all three commits individually.


gammons:
Both races are real — I verified them the way I'd want them verified, and the fix closes the connect-goroutine and WebSocket-goroutine sides completely. But it has the same shape of hole as #147: one access site bypasses the new store.

**Confirmed real:**

1. `activeTeamID` — written from a tea.Cmd goroutine (the workspace switcher) and connect goroutines, read from the UI goroutine (savers, `setStatus`) and every WS goroutine (`isActive`). Deleting it in favor of `router.Active()` rather than synchronizing it is the right call — it was a duplicate of `router.Active().TeamID`.
2. `cfg.Workspaces` — `SaveTheme`/`SaveSidebarWidth` read-modify-write on the UI goroutine while `connectWorkspace` (`ResolveTheme`/`ResolveWidth`, `buildChannelItem` section lookups) and `h.cfg` in `OnConversationOpened` read the same aliased map from connect and WS goroutines. Fatal concurrent map access, matching your track record on #202/#205/#208.

The `Snapshot()` clone for `rtmEventHandler.cfg` and the mutex on the store are right, and I verified the tests are genuinely red→green — reverting the mutex + map-clone in `workspace_config_store.go` makes `TestWorkspaceConfigStore_ConcurrentSaveAndSnapshot` fail with both a `-race` report and a runtime fatal.

**The hole: the workspace switcher closure still reads the outer `cfg`.**

At `cmd/slk/main.go:2005-2006`:

```go
Theme: cfg.ResolveTheme(teamID),
SidebarWidth: cfg.ResolveWidth(teamID),
```

This runs on a tea.Cmd goroutine. `newWorkspaceConfigStore(cfg)` copies the struct but aliases the same `Workspaces` map, so when `[workspaces.*]` blocks exist, `SaveTheme`'s mutex-guarded write (UI goroutine) races this unguarded map range (cmd goroutine). The store's mutex can't protect a read that never goes through the store.

There's a second, related consequence: when `cfg.Workspaces` is nil at boot (any user without `[workspaces.*]` blocks), `tomlKeyFor` allocates a **private** map on the store, so the outer `cfg.Workspaces` stays nil forever — a workspace-scoped theme saved this session is silently lost on workspace switch (falls back to global, self-heals on restart). Pre-PR, `saveTheme` allocated on the shared `cfg`, so the switcher saw it.

One change fixes both: route the switcher's `ResolveTheme`/`ResolveWidth` through the store (e.g. `workspacesStore.Snapshot()` or dedicated getters).

**Smaller notes:**

- Your PR description still describes the intermediate `activeTeam{id string}` design, but the final commit deletes it in favor of the router. The code is better than the description — update it.
- `TestWorkspaceRouter_ConcurrentSetAndActive`/`ConcurrentAddAndByID` pin already-safe code; your commit message says so honestly, which I appreciate.
- The switcher comment at `main.go:1983-1984` ("callbacks read router.Active()") now papers over the closure itself still reading outer `cfg`.

Also verified: no deadlock risk (locks never held across callbacks), snapshot staleness benign for its actual uses, build/vet/gofmt/race all green.

Fix the switcher read and this merges.

laraibg786:
Thanks for the thorough review, addressed the hole and the two notes.

1. PR description has been updated to lead with the final state (no `activeTeamID`, no `activeTeam`) instead of dwelling on the transient scaffolding type, and now documents all three fixes under commit 3.
2. `main.go`'s workspace-switcher callback was still reading the outer `cfg` directly for `Theme`/`SidebarWidth` instead of `workspacesStore`. Routed both through `workspacesStore.Snapshot()`, same pattern the connect path already used, closes both the race against `SaveTheme`/`SaveSidebarWidth` and the silent-loss case when `cfg.Workspaces` started nil at boot. Folded into commit 3 rather than tacking on a fourth commit, since it's the same store getting the same kind of fix.

All of your review concerns have been addressed. Please review and merge this PR whenever you have time.

PR #226, opened 2026-09-17
fix(statusbar): render spacers without StatusBar's padding

### LINKED ISSUE #181: View() renders 6 columns wider than the terminal width, wrapping the status row

laraibg786:
styles.StatusBar carries Padding(0, 1), so every spacer rendered through it came out 2 columns wider than the gap arithmetic assumed: +6 with the help hint, +4 without. Nothing clipped the result, so View() returned a frame wider than the terminal and the status row wrapped, scrolling the top line away.

Render the filler and the right gutter through a background-only style instead. Re-blessed the goldens and set goldenStatusOverflow to 0, which also drops the overlay exception in the width guard.

Fixes #181


gammons:
@laraibg786 do you have the same bug as me?  on the bottom right, the "Connected" string is cut off.  This is also on `main` so it's not something your PR introduced.  but probably is a related fix we could throw in this PR possibly.

image

laraibg786:
Yes we are fixing this bug where connected is cut off in between.

gammons:
thanks @laraibg786 !

Issue #180, opened 2026-09-05
Feature proposal: export one channel's messages and threads to Markdown


jftuga:
Could slk support exporting one channel's messages and threaded replies for a date range? This would extend the usefulness of saving individual threads with `S`.

```sh
slk export \
  --workspace example-workspace \
  --channel project_alpha \
  --since 2026-04-01 \
  --until 2026-07-01 \
  --timezone America/New_York \
  --output ./project-alpha-export \
  --overlap 14
```

- Use existing workspace credentials to fetch one channel directly, without launching the TUI.
- Save each conversation, including standalone messages, to a Markdown file with an `index.md` linking them together.
- Treat `--since` as inclusive and `--until` as exclusive in the selected timezone.
- Let `--overlap N` add N calendar days before `--since` and after `--until` (default: zero).
- Include replies within the expanded interval even when their parent is older, retaining that parent as context.

Would you accept a PR for this feature? I'd like to confirm the scope and CLI approach before starting implementation.

jftuga:
@gammons what are your thoughts on this?

jftuga:
I have created: https://github.com/gammons/slk/pull/227

PR #227, opened 2026-09-17
feat: export a channel's messages and thread to Markdown

jftuga:
## Summary

`S` saves one open thread. This adds the bulk counterpart: `slk export` writes one channel's messages and threaded replies for a date range to Markdown, using the stored workspace credentials and without launching the TUI.

```sh
slk export \
  --workspace example-workspace \
  --channel project_alpha \
  --since 2026-04-01 \
  --until 2026-07-01 \
  --timezone America/New_York \
  --output ./project-alpha-export \
  --overlap 14
```

## Behavior

- `--since` is inclusive and `--until` exclusive, both read as midnight in `--timezone` (default local). `--until` defaults to tomorrow.
- `--overlap N` widens the range by N calendar days on each side (DST-safe).
- Each top-level message becomes one Markdown file holding its in-range replies, and `index.md` links them by day. Bodies reuse `export.ThreadToMarkdown`, so output matches a thread saved with `S`.
- A reply inside the range is exported even when its parent is older; that parent is kept and labelled as context.
- `--output` must be empty or absent, checked before fetching. 
- `--workspace` is optional with a single workspace
- `--channel` takes a name (starting with **#**) or an ID.

## Design notes

- Finding older threads with in-range replies means walking all of `conversations.history` before the range and keeping parents whose `latest_reply` reaches it. Slack has no "threads active in this period" query, and a bounded lookback or `search.messages` would silently miss replies. The cost is one request per 200 earlier messages, so a first export of a long-lived channel is slow.
- New `WalkHistory` and `GetRepliesBetween` in `internal/slack/history_range.go` stream pages and wait out rate limits. Existing fetch methods are untouched, so no TUI path changes behaviour.
- Flags use the standard library `flag` package, matching the existing hand-rolled CLI rather than adding a framework dependency.
- Design doc: `docs/superpowers/specs/2026-09-17-channel-export-design.md`. `AGENTS.md` and `wiki/Features.md` are updated.

## Testing

- Unit tests cover window edges, overlap across DST, file and index output, paging, rate-limit retry, cancellation, conversation selection around the range, name resolution and flag parsing.
- `go build ./...`, `go vet ./...`, `go test ./... -race`, `gofmt -l .` and `golangci-lint run` all pass.
- Also tested by hand against a live workspace.

## Known limitations

- Messages carried only in Block Kit blocks or legacy attachments (many bot posts) export with an empty body. This is existing `ThreadToMarkdown` behaviour, affects `S` too, and belongs in a separate change.
- System messages are exported like any other message. Files are linked, not downloaded.

## Disclaimer

This PR was vibe-coded with Claude Fable 5.1 at xhigh effort. It was also tested by hand.

Related Issue: https://github.com/gammons/slk/issues/180


teluka:
that would be super helpful feature :+1:

gammons:
Thanks for this — it closes #180, and it's the most carefully built PR in the queue right now. I verified every claim in the description rather than taking them on trust, and they all hold, on your branch merged with current `main`:

| check | result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l .` | empty |
| `go test ./cmd/slk/... ./internal/export/... ./internal/slack/ -race` | pass |
| `golangci-lint run` | 0 issues |

**It survives the Phase 1 split.** #241 landed after you opened this and moved `messageAuthor` and `resolveUserCached` out of `main.go` into `users.go`. Your branch still merges cleanly, builds, and passes — including the new `main_scope_test.go`, which fails any PR that adds a declaration to `main.go`. Your two new files are already topical and land exactly where that test wants them. Nothing for you to do; noting it so you know it was checked.

**On the tests: they have real teeth.** I don't usually say that, so here is the evidence. I ran eight mutations against the core logic and every one was caught:

| mutation | caught by |
|---|---|
| `Contains` end bound made inclusive | `TestWindow_Contains`, `TestWindow_ContainsUsesOverlap` |
| drop the cross-page reply dedup | `TestWindowReplies_DropsParentDuplicatesAndOutOfWindow` |
| `repliesReachWindow` ignores `latest_reply` | `TestCollectConversations_FetchesRepliesOnlyWhereTheyCanMatter` |
| `isThreadReply` always false | `TestCollectConversations_SelectsByWindow` |
| `stampMessage` uses UTC, not the window zone | `TestWriteChannel_TimesFollowTheWindowTimezone` +3 |
| overlap widens only the start | `TestNewWindow_OverlapAddsCalendarDaysAcrossDST` +3 |
| keep context parents with no replies | `TestCollectConversations_SelectsByWindow` +1 |

The `WalkHistory`-before-the-window design is the right call, and the description is honest about what it costs. Agreed that a bounded lookback would silently miss replies, which is worse than being slow.

Four things, none of them blocking.

**1. Two unparseable timestamps overwrite each other, and the index says otherwise.**

`conversationFilename` (`internal/export/channel.go:167-174`) falls back to `"unknown"` for both halves of the name when `TimeFromTS` fails, so every such conversation resolves to the same file. Reproduced on your branch:

```
2 conversations -> files: [index.md unknown-unknown.md]
surviving content contains FIRST=false SECOND=true

index.md:
- Conversations: 2
- [ a: FIRST](unknown-unknown.md)     <- this file no longer exists as written
- [ b: SECOND](unknown-unknown.md)
```

The first conversation is gone, and the index reports a successful export of both. You already handle unparseable timestamps deliberately one function up — `stampMessage` (`channel.go:154-162`) blanks the display fields rather than inventing a time — so the case was clearly on your mind; it just isn't carried through to the filename. `TestWriteChannel_SameSecondParentsGetDistinctFiles` pins the collision you did think of. Disambiguating with the conversation's index, or falling back to the raw TS, would close it. Slack timestamps are always well-formed in practice, so this is robustness rather than a live bug — but silent data loss with a confirming index is the wrong failure mode to leave in.

(Same root cause, cosmetic: a blank `Timestamp` leaves the index label starting with a space — `- [ a: FIRST]`.)

**2. The cancellation plumbing cannot fire in production.**

`exportChannel` uses `ctx := context.Background()` (`cmd/slk/export_command.go:184`), and `os/signal` is imported nowhere in the repository — I grepped the whole tree, including after the Phase 1 split. So the `ctx.Err()` checks at the top of `WalkHistory` and `GetRepliesBetween`, the `ctx.Done()` select in `WaitOutRateLimit`, and the two tests covering them (`TestWalkHistory_RateLimitWaitHonorsCancellation`, `TestResolveExportNames_CancelledDuringRateLimitWait`) are all unreachable once the binary is running.

That matters more here than anywhere else in slk: this is the one command the description says can run for many minutes on a long-lived channel, and `WaitOutRateLimit` will sleep 30s at a time with no retry ceiling. Ctrl-C still kills the process, so nothing hangs — but the careful ctx-aware code is decoration until `signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)` replaces that line. That's a one-line change that makes an already-written feature real.

**3. A failed export poisons its own retry.**

`WriteChannel` writes conversation files in a loop and only fails partway through on a write error (`channel.go:88-90`). The directory is then non-empty, so the retry hits `ErrOutputNotEmpty` — and because `DefaultChannelDir` is deterministic, re-running the identical command fails until the user deletes the directory by hand. Writing to a sibling temp dir and renaming on success would make the operation atomic. The pre-flight `EnsureEmptyDir` at `export_command.go:204` is a good touch and worth keeping either way.

**4. `fetchUserName` cannot distinguish "no such user" from "network down".**

`export_collect.go:279-293` returns `("", nil)` for every non-rate-limit error. The doc comment frames this as bot IDs and deleted users, which is right for a 404 — but a DNS failure or a 500 takes the same path. After a 40-minute fetch, names silently degrade to raw `U…` IDs in the output with nothing on stderr. Since the export is already tolerant here, a warning line would be enough.

**One coordination note.** This adds a fourth caller of `resolveUserCached` (`export_collect.go:256`); #147 proposes deleting that function outright. Whichever lands second needs a small fixup — flagging it so it isn't a surprise. Post-#241 the function now lives at `cmd/slk/users.go:47`.

Marking this **needs minor changes**. The design, the tests and the docs are all where I want them; it's the four items above and I'm happy to merge.

jftuga:
Thanks for the thorough review. All four items are addressed, one commit each, on top of 2affd6f. None of the commits touch `main.go`, so the Phase 1 `main_scope_test.go` is unaffected, and `resolveUserCached` is still called from the same one place in `export_collect.go` for whichever of this and #147 lands second.

### 1. Two unparseable timestamps overwrote each other (90a1539)

`WriteChannel` now passes every filename through a new `uniqueFilename`, which appends a counter before the extension to any name the export has already used and records the result. Two conversations whose TS does not parse now land in `unknown-unknown.md` and `unknown-unknown-2.md`, and the index links each to the file that actually holds it. The guarantee is structural rather than tied to timestamp shape: no two conversations in one export can share a file, whatever their TS looks like.

The same blank stamp also produced the leading space in the index label, an empty `## ` day heading, and a title ending in ` - `. Those now omit the time, read `## Unknown date`, and omit the date suffix respectively.

Tests: `TestWriteChannel_UnparseableTimestampsGetDistinctFiles` reproduces your two-conversation case and asserts both files survive, both are linked from the index, the labels start with the author, and the heading is labelled. `TestUniqueFilename` pins the counter sequence.

### 2. Cancellation could not fire (e91e384)

`exportChannel` now builds its context with `signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)`, so the existing `ctx` checks in `WalkHistory`, `GetRepliesBetween` and `WaitOutRateLimit` are reachable in the binary. One addition beyond the one-liner: `context.AfterFunc(ctx, stop)` restores default signal handling as soon as the first signal cancels `ctx`. Without it, `NotifyContext` keeps swallowing SIGINT until the function returns, so a second Ctrl-C would do nothing during a step that does not watch `ctx` (connecting, the per-user profile lookups, writing the files). With it, the first Ctrl-C cancels the fetch cleanly and a second one still kills the process.

Verified outside the test suite with a small program using the same three lines: the first self-sent SIGINT cancels `ctx` with `context.Canceled`, the second terminates the process with exit 130. The binary cross-builds for linux and windows.

### 3. A failed export poisoned its own retry (d00b489)

`WriteChannel` now writes into a sibling `<dir>.partial-*` directory created with `os.MkdirTemp`, and only once every conversation file and the index are written does it move them into the output directory. A write failure removes the staging directory and returns the error, leaving the output directory exactly as it was found: empty, and ready for an identical re-run. The pre-flight `EnsureEmptyDir` in `exportChannel` is unchanged.

One deliberate deviation from the suggestion: the staging directory is moved into place file by file, not renamed over the output directory. I tried the rename first. macOS refuses `os.Rename` onto an existing directory even when it is empty, so the portable form would be remove-then-rename, and that swaps the inode out from under any shell whose working directory is the output directory (`slk export --output .` from an empty directory is a plausible invocation). The per-file moves are each atomic, stay on one filesystem, and keep the property that matters here: nothing touches the output directory until the whole export has been written.

Tests: `TestWriteChannel_FailedWriteLeavesDirEmptyForRetry` forces a mid-loop write failure (a 300-digit sequence number produces a filename longer than any filesystem accepts) after one file has already been written, then asserts the output directory is empty, no staging directory is left beside it, and a retry into the same directory succeeds. `TestWriteChannel_IntoCurrentDirectory` exports to `.` and confirms the index is readable relative to the working directory afterwards. The existing one-file-per-conversation test additionally asserts nothing is left beside the output on success. I also mutation-checked the cleanup: dropping the `os.RemoveAll` fails the retry test.

### 4. `fetchUserName` hid network errors (11f0f6c)

`fetchUserName` now returns the lookup's error instead of `("", nil)`. `resolveExportNames` takes a `warn io.Writer` (stderr in `exportChannel`) and prints one line per ID it could not resolve, quoting the error, so `user_not_found` and a DNS failure read differently:

```
Warning: could not resolve user U404, keeping the ID: getting user info: user_not_found
Warning: could not resolve user U1, keeping the ID: getting user info: dial tcp: lookup slack.com: no such host
```

The export is still tolerant: the ID keeps its raw form and the run completes. Only a cancelled context aborts it, which the existing `TestResolveExportNames_CancelledDuringRateLimitWait` still covers and now also asserts is not reported as a lookup warning.

Tests: new `TestResolveExportNames_WarnsAndKeepsIDWhenLookupFails`; `TestResolveExportNames_AuthorsAndMentions` now asserts the exact warning for its `U404` case.

The design doc's output-directory and name-resolution bullets are updated in the commits that change them.

### Verification on the branch

| check | result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l .` | empty |
| `go test ./... -race -count=1` | pass, 57 packages |
| `golangci-lint run` | 0 issues |

PR #228, opened 2026-09-18
chore: add self-hosted renovate for dependency updates

laraibg786:
## Why Renovate over Dependabot

Three things this repo needs that Dependabot can't express: a release-age gate (`minimumReleaseAge`, so nothing lands minutes after publish), per-class automerge in config rather than a bespoke workflow, and converting the four action tags to pinned digests. `release.yml` publishes signed binaries to Homebrew and AUR, so floating action tags there are the gap worth closing first.

## What automerges

| class | automerge |
|---|---|
| GitHub Actions pin/digest/minor/patch | yes, grouped |
| any Go module | no |
| `golang.org/x/**`, `modernc.org/**` | no, grouped |
| `charm.land/**`, `github.com/charmbracelet/**` | no, grouped |
| any major | no |
| `flake.lock` refresh | no |

Only Actions automerge, gated on `ci`. Everything else opens a PR and waits.

Go modules are held back because `flake.nix` pins a `vendorHash` derived from `go.mod`/`go.sum` — any bump invalidates it and breaks `nix build`, and nothing in CI currently catches that. Once #NNN (the nix build job) lands, this can relax to automerging patches.

Charm stays manual for a second reason: `internal/ui/golden_test.go` renders through bubbletea and `x/ansi`, so a bump shifts the eight `.ansi` goldens and needs them re-blessed by hand.

Later `packageRules` win in Renovate, so the charm rule sits last.

## Nix

The nix manager is beta and opt-in, so it's enabled explicitly with `lockFileMaintenance`. Both `flake.nix` inputs are branch refs, so there's no version to bump there — `flake.lock`'s pinned revs are the update surface. `flake-utils` is currently pinned to a rev from 2024-11 and `systems` to 2023-04.

## golangci-lint

The version was pinned inline as a step input, which no bot can see. It's now a workflow-level `env:` with a marker comment, so the regex manager tracks it:

```yaml
env:
  # v2.13.0+ is required for go1.27 ...
  # renovate: datasource=github-releases depName=golangci/golangci-lint
  GOLANGCI_LINT_VERSION: v2.13.1
```

The marker is needed rather than matching a bare `version:` — that would also match goreleaser's `version: "~> v2"` in `release.yml`. Pattern matches once across the three workflows.


laraibg786:
@gammons please review this. please comment if you want to do anything different here.
This PR requires some setup from the repo admin.

## Setup

Nothing happens until `RENOVATE_TOKEN` exists, so this can merge and sit dormant. `GITHUB_TOKEN` won't work: PRs it opens don't trigger CI, and it can't modify `.github/workflows/`.

Fine-grained scopes are Contents, Issues, Pull requests, Workflows, Commit statuses, Dependabot alerts — all read/write, scoped to this repo.

once tokens are generated you can paste
```sh
gh secret set RENOVATE_TOKEN --repo gammons/slk --app actions   # paste at the prompt
```

gammons:
this makes sense to get done.  thanks!

PR #229, opened 2026-09-18
chore(CI): add NIX build job to make sure that dependency or other updates do not break NIX

laraibg786:
## Problem

`flake.nix` pins `vendorHash`, derived from `go.mod`/`go.sum`. Any dependency change invalidates it and breaks `nix build`, but nothing in CI builds the flake — so it surfaces for whoever runs `nix build` next, at a commit unrelated to the one that caused it. The repo ships a flake as an install path and nothing currently proves it works.

## Note for Maintainer

If you want the workflow to trigger on the commit by the bot, you have to add `NIX_HASH_PUSH_TOKEN` secret in the repo with content(read,write) on the repo. The commit is pushed anyways but commits pushed by `github.token` do not trigger workflow or commit statuses.


gammons:
The approach is right, and two details show real care: pushing the hash fix only when a token that can retrigger CI is configured (a `GITHUB_TOKEN` push would leave the head SHA with no checks at all — most people miss that), and proving the refreshed hash is the *only* problem by rebuilding with it before pushing.

**One thing to fix before merge: the fork-PR path.** `nix-vendor-hash.yml` runs on `pull_request` and checks out `head.repo` with `NIX_HASH_PUSH_TOKEN || github.token`. Once the PAT exists, a fork PR gets a workflow that runs `nix build` — arbitrary build code from the PR's `flake.nix` — in an environment where a real PAT with `contents: write` is present (checked out via `token:` and exported in env). A malicious fork PR could exfiltrate it. The job-level `contents: write` is forced read-only for forks when only `github.token` is in play, but the PAT changes that.

Gate the token use to same-repo PRs:

```yaml
if: github.event.pull_request.head.repo.full_name == github.repository
```

on the vendor-hash job (or on the steps that use the token). The build job in `ci.yml` is fine as-is — it runs without secrets.

With that guard, I'm happy to merge this and add the `NIX_HASH_PUSH_TOKEN` secret.

laraibg786:
Sorry for overlooking the vulnerability in my workflow making it prone to supply chain attacks by malicious forks. i have updated the workflow and added the check to secret retrieval at every place. should be good now. please let me know if you have any further suggestions on this.

laraibg786:
@gammons sorry for bugging you again but could you please take a look and review the blockers in merging this PR. If nothing is outstanding we should be able to merge this.

gammons:
sorry about that @laraibg786 !

laraibg786:
no worries. make sure to update the pr labels as well.

gammons:
Labels updated — swapped `needs minor changes` for `ready to merge` to match how this actually landed. Sorry for leaving the stale one on after merging.

For the record, since this merged before I wrote the follow-up: the fork-PR gate is exactly what I was after. `PUSH_TOKEN` is gated on `head.repo.full_name == github.repository` in both the job `env` and the checkout `token:`, and the push step additionally requires `env.PUSH_TOKEN != ''`, so a fork PR gets the patch artifact and never sees the PAT. Belt and braces, given `pull_request` already withholds secrets from forks — and worth keeping precisely for the day someone reaches for `pull_request_target`.

One note for whoever touches this next: `nix-vendor-hash.yml` hasn't executed yet. Its `paths:` filter is `go.mod`/`go.sum`/`flake.nix`/`flake.lock`, and this PR only touched workflow files, so the 128 lines of it are still untested. #233 bumps `modernc.org/sqlite` and will be its first real run — I've asked there for a rebase so it triggers. It fails safe if the hash can't be parsed, so the downside is a red check rather than a bad push.

Thanks for the ping about the labels, and for the patience on the review turnaround.

PR #230, opened 2026-09-18
fix(membership): don't mark everyone "not in channel" on a cold cache

landonforshage:
## Problem

In a four-person group DM on an Enterprise Grid workspace, typing `@Britt` offered "Brittany (not in channel)". Every candidate in every conversation on that workspace gets the same label.

## Cause

Verified on 2026-09-18 against the running session's cache and a live API call:

1. The workspace refuses `conversations.members` (`enterprise_is_restricted`, on both the Grid host and `slack.com`). The cache has no `channel_members` or `channel_membership_meta` rows for any of its conversations; the other four workspaces in the same cache have rows.
2. On a cold cache, `EnsureFresh` still pushed a snapshot: `loadIntoMemory` stored an empty set and `pushSnapshot` sent it.
3. `compose.SetChannelMembership` treats any pushed list as loaded, so `rebuildMentionUsers` checked every user against an empty set.

The spec already covers a failed fetch: the picker "silently degrades to 'no membership data' mode (everyone shown without labels)" (`docs/superpowers/specs/2026-05-20-mention-picker-channel-membership-design.md`, Failure Modes). The empty push meant that mode never applied. It came from the lifecycle step "Stale or missing: push current membership to UI immediately", which with nothing cached pushes an empty list. On other workspaces the same push makes the labels flash until the first fetch lands.

## Fix

A channel without a `channel_membership_meta` row has never been fully fetched, so the manager now treats its list as unknown: `loadIntoMemory` leaves it out of the map and `pushSnapshot` doesn't push it. `ForceStale` zeroes the timestamp but keeps the row, so a force-staled channel is still known. The `pushSnapshot` check also stops `ApplyLeave` from pushing an empty list for a channel that isn't loaded.

`ApplyJoin` used to create a set holding only the joiner whenever the channel wasn't loaded. On a never-fetched channel that pushed the same wrong labels. On a channel cached but not yet opened, the new set hid the cached list for up to the 24h TTL. It now loads the cached list first and adds to it only if one exists. Loading first also covers a join that lands while `EnsureFresh` is loading the same channel.

## Why not fix it in the compose model

Treating an empty list as "not loaded" there is one line, but the compose model can't tell an empty list from an unknown one, and it would still accept the joiner-only list. Only the manager knows whether a list came from a fetch.

## Tests

In `internal/slack/membership/manager_test.go`:

- `TestEnsureFreshNeverFetchedPushesNothing`: cold cache, fetch refused, nothing pushed.
- `TestApplyJoinOnNeverFetchedChannelPushesNothing`: a join on a never-fetched channel pushes nothing.
- `TestApplyJoinKeepsCachedMembersNotYetLoaded`: a join on a cached, unopened channel pushes one list with all three members.

Five existing tests used the cold-cache empty push as a sync point; those receives are removed.

Red with each guard removed:

- The meta check in `loadIntoMemory`, or the unknown check in `pushSnapshot`: `TestEnsureFreshNeverFetchedPushesNothing` and `TestApplyJoinOnNeverFetchedChannelPushesNothing` fail. So do three existing tests (`TestEnsureFreshCacheMissTriggersFetch`, `TestBackgroundFetchDoesNotResolveEveryMember`, `TestBackgroundFetchRetriesAfterBackoffExpiry`), with messages about the fetch, because the stray empty push arrives where they expect the fetch's.
- Restoring the old create-on-nil in `ApplyJoin`: `_OnNeverFetchedChannelPushesNothing` fails.
- Dropping the load in `ApplyJoin`: `_KeepsCachedMembersNotYetLoaded` fails. The concurrent-load case has no deterministic test, because that would need a hook inside `loadIntoMemory`.

Build, vet, `go test ./... -race`, `gofmt -l` and `golangci-lint run` (v2.13.1) are clean.

**Field check (macOS, same workspace, patched build):** in the same group DM, `@Britt` now offers Brittany with no label.

## Noted but not changed

- Where `conversations.members` is refused, each channel switch after the 5-minute `FailureBackoff` sends another request that will fail the same way. On this workspace `conversations.info` does return `members` for group DMs, so that could give real in-channel ordering there. It would be a new fallback, not part of this fix.
- From the Codex review: a join that lands between `backgroundFetch`'s `ReplaceChannelMembers` and its in-memory install is overwritten by the fetch's older set, and it stays that way until the next full fetch. This existed before the PR, and the window is microseconds.

## AI assistance

Written with Claude (Opus 5) driving the investigation, code and tests after I hit the bug in a live session. The cause was confirmed against the running session's cache and a live `conversations.members` response, not inferred from the code. Codex and Claude (Fable 5.1) reviewed the diff. Codex's second pass caught a lost-join race in a simplification of `ApplyJoin` that I then dropped, and Fable caught an inaccurate test comment. I've read the diff and will defend it in review.


gammons:
Verified and this is right. Merging.

The root cause chain checks out: on a workspace that refuses `conversations.members` (`enterprise_is_restricted`), there are no `channel_members`/`channel_membership_meta` rows at all. `loadIntoMemory` installed an empty set anyway, `pushSnapshot` pushed it, and `SetChannelMembership` treats any pushed list as loaded — so the mention picker checked every user against an empty set and everyone got "(not in channel)".

The fix is the right shape: without a meta row, cached rows are join deltas, not a member list, so nothing gets installed — and unknown channels simply aren't pushed, leaving the picker in its not-loaded state (which degrades gracefully per the spec) instead of confidently wrong. The `ApplyJoin` load-first fix closes the adjacent race where a join landing during `EnsureFresh`'s load would be dropped from the set being installed.

I verified the tests have teeth: reverting both the meta-row check and the push guard fails five tests, including your two new ones (`TestEnsureFreshNeverFetchedPushesNothing`, `TestApplyJoinOnNeverFetchedChannelPushesNothing`). The existing tests that relied on `EnsureFresh`'s synchronous empty push as a barrier are correctly updated rather than papered over.

Build, gofmt, and `go test ./internal/slack/membership/` all clean. Merging.

Issue #188, opened 2026-09-07
Four unreachable code blocks in the mode handlers


gammons:
Phase 0's characterization pass over the 16 mode handlers turned up four blocks
that cannot execute. None is user-visible; each was verified by tracing the only
caller. Filing together because the fix is the same in every case — delete, or
explain why it stays — and because Phase 4 (modal chrome substrate) will be
touching this code.

Low priority. Listed roughly in order of how confidently they can be removed.

### 1. `mode_confirm.go:19-20` — Esc normalisation has no effect

```go
switch msg.Key().Code {
case tea.KeyEscape:
	keyStr = "esc"
case tea.KeyEnter:
	keyStr = "enter"
}
res := a.confirmPrompt.HandleKey(keyStr)
```

`confirmprompt.HandleKey` (`internal/ui/confirmprompt/model.go:76-89`) switches
on `"y", "Y", "enter"` and lets everything else fall to a `default` that
cancels. There is no `"esc"` case. So rewriting Escape to `"esc"` changes
nothing: both the normalised and un-normalised forms hit the same default.

The `KeyEnter` arm of the same switch **is** load-bearing (it is what makes
`shift+enter` confirm) — only the `KeyEscape` arm is inert.

### 2. `mode_workspace_search.go:37-39` — the `!ok` branch is unreachable

```go
case searchresults.ActionSelect:
	item, ok := a.searchResults.Selected()
	a.searchResults.Close()
	a.SetMode(ModeNormal)
	if !ok {
		return nil
	}
```

`searchresults.HandleKey` returns `ActionSelect` **only** inside
`if _, ok := m.Selected(); ok` (`internal/ui/searchresults/model.go:149-153`).
`Selected()` is a pure value-receiver read, and the handler repeats it at `:34`
before `Close()`, so `ok` cannot be false here.

This is the sole reason `handleWorkspaceSearchMode` stops at 96% statement
coverage rather than 100%.

### 3. `mode_normal.go:196-198` and `:206-208` — `PageDown` / `HalfPageDown` inner returns

```go
case key.Matches(msg, a.keys.PageDown):
	if cmd := a.scrollFocusedPanel(a.pageSize()); cmd != nil {
		return cmd
	}
```

`scrollFocusedPanel` (`internal/ui/app.go:1461-1498`) returns a non-nil `tea.Cmd`
on exactly one path, `:1486`, which requires `PanelMessages` **and** a negative
[...]

Issue #236, opened 2026-09-20
RFC: Self-contained UI components in `internal/bubbles/`


laraibg786:
Design for turning `internal/ui/<widget>/` into self-contained, domain-agnostic Bubble Tea components under `internal/bubbles/<widget>/`: each owning its own `Styles`, `KeyMap` and size, and taking the callables it needs as explicit parameters to `New` so it can issue its own commands instead of `App` doing that for it.

This RFC does **not** propose publishing a library. `internal/bubbles/` is the destination; whether any of it later moves to `pkg/` is a separate decision taken after the conventions have proven themselves in-tree.

## Problem

slk has 29 component packages and a 13,561-line app layer. The components look self-contained and are not. Measured at `4d4365b` from each package's non-test `GoFiles`:

**The theme is a process-global mutable singleton.** `internal/ui/styles` exports 20-odd `color.Color` vars plus `Apply()` that mutates them, and components read it **394 times across 23 of the 29 packages** — `messages` 72, `thread` 49, `statusbar` 29, `sidebar` 26. A component that reaches into a global for its colours cannot be instantiated twice with different looks, which a split view already wants, and cannot be reasoned about in isolation.

**No component participates in the Bubble Tea contract.** Zero implement `tea.Model`, zero have `Init()`, exactly one (`compose`) has `Update`. Every `View` takes geometry — `View(width int)`, `View(height, width int)`, `View(termWidth, termHeight int)` — so no component owns its own size. Construction is ad-hoc: six different `Open(...)` signatures, no common shape. Keys arrive as pre-normalized strings, losing modifiers, and `HandleKey` has nine distinct return types, which is why no modal interface is declarable.

**Components cannot act; they can only be driven.** No component holds a service port — every one of `core`'s 14 service interfaces is used only by `App`. That sounds like clean layering, and at the I/O boundary it is, but it means a widget cannot express *"the user asked me to mark this read"* as anything but a return value for `App` to interpret. The consequence is measurable in the reducers: of 30 non-test `reducer_*`/`mode_*` files, **12 touch exactly one widget** and are largely couriering between it and one service. (Four touch no widget at all; the remaining 14 fan out to two or more and are a different problem — see Non-goals.)

[...]

PR #237, opened 2026-09-20
refactor(bubbles): extract confirmprompt as the first self-contained component

laraibg786:
## What changes

`internal/ui/confirmprompt` becomes `internal/bubbles/confirmprompt`, rebuilt to the Bubbles v2 shape:

- `KeyMap` + `DefaultKeyMap()` and `Styles` + `DefaultStyles(isDark)`, so the host overrides instead of the component reading the theme global. The package no longer imports `internal/ui/styles`.
- `Update(tea.Msg) (Model, tea.Cmd)` replaces `HandleKey(string)`, with a concrete return type. Modifiers are ignored on non-printable keys, so a stray shift on Enter still confirms — the old code did this with a `Code` switch in the caller.
- `View() string` and `SetWidth()`: the model owns its size and renders only its own box.
- `New(opts ...Option)` with `WithStyles` / `WithKeyMap` / `WithWidth`, matching `table`, `viewport`, `progress`, `spinner`, `paginator` and `timer` in bubbles v2. `Option` is `func(*Model)` applied after defaults, runtime setters stay alongside, and there is no chaining builder because bubbles has none.
- `Styles()` / `SetStyles()` accessors, as `table`, `textinput` and `textarea` do in v2; `KeyMap` stays an exported field, as on `table` and `list`.
- The prompt's one dependency is a named `ConfirmFunc func() tea.Msg` rather than an inline signature, so `Open` states what it takes.
- Footer help text derives from `KeyMap` instead of a literal.

The package's entire import set is `bubbletea`, `bubbles/key`, `lipgloss`, `reflow/truncate` and `strings` — no slk, no Slack, no sibling widget, and nothing that reads the theme global.

## Behaviour

Two deliberate differences, both recorded:

**`SetSize` becomes `SetWidth`.** The height argument wrote a field nothing read once compositing moved to the host. Box height follows content and the caller centres it, so terminal height was never the model's business.

**The `esc` arm of `mode_confirm.go`'s `Code` switch is dropped.** `mode_confirm_test.go` records it as having no effect on behaviour — gammons/slk#188 item 1. The `shift+enter` arm *was* load-bearing and is preserved inside the component.

Everything else is byte-identical, verified against captured output of the old implementation across five terminal sizes plus the composited overlay and box geometry.

## Verification

- `go test ./internal/bubbles/... ./internal/ui/` green; the 8 App-level goldens are unchanged.
- `golangci-lint run` — 0 issues.
- The component's own tests run in 3ms with no `App`, which is the point: today a confirm-prompt change is gated on the 6s `internal/ui` suite.


laraibg786:
This PR is a pilot refactor based on the RFC #236 and would be changed depending on the acceptance/updates on the RFC.

gammons (commented):
Reviewed against a local checkout at 75606ce. Re-ran everything: `go build ./...`,
`go test ./... -race`, `go vet ./...`, `gofmt -l .`, `golangci-lint run` — all
clean, 0 lint issues, the 8 goldens are byte-identical. I also diffed the old
`renderBox` against the new `View` line by line: the 35%/40/60 clamp, the `-4`
chrome, the `> ` prefix, the `\n\n` joins, `truncate.StringWithTail` and the
footer all match, and the `KeyMap`-derived footer reproduces the old literal
exactly under `DefaultKeyMap()`. The `esc`-arm removal really is inert — the old
`default` cancelled on anything, so `"esc"` and `"escape"` were already the same
path.

This is the right shape and a good template. Three things I'd want fixed before
it merges, because the whole point of a first migration is that 28 PRs copy it.

---

### 1. Styles are snapshotted at `Open`, and the justification doesn't hold

`confirm.go:18-20` says the prompt "cannot be visible across a theme change"
because `ModeConfirm` and `ModeThemeSwitcher` are mutually exclusive. That only
accounts for one of the three `styles.Apply` sites. The other two are in
`reducer_workspace.go` — `:200` (`WorkspaceReadyMsg`, per-workspace theme on
initial activation) and `:350` (`WorkspaceSwitchedMsg`) — and reducers run from
`dispatchReducers` with no mode gate. Press `q` during startup while a workspace
is still connecting and the quit prompt renders in the previous theme's colours
until it closes.

Old code re-read `styles.Background` every frame, so it followed. This is a real
regression. Repro, which passes on `main` and fails on the branch:

```go
func TestConfirmPromptFollowsThemeChange(t *testing.T) {
	styles.Apply("dracula", core.Theme{})
	a := newTestApp(t, withSize(120, 40))
	a.openConfirmPrompt("Quit slk?", "All workspace connections will close.", nil)

	before := a.confirmPrompt.View()
	styles.Apply("nord", core.Theme{}) // what reducer_workspace.go:350 does
	if a.confirmPrompt.View() == before {
		t.Error("prompt did not follow the theme change")
	}
}
```

The fix is the push you already have, at the right sites: add
`a.confirmPrompt.SetStyles(confirmPromptStyles())` next to the existing
`a.compose.RefreshStyles()` calls at all three `styles.Apply` locations. #236
says this itself — *"the snapshot mechanism must land before any component stops
reading the global."* Worth landing that mechanism generally here rather than
per-widget later.

### 2. `reassertBase` is a verbatim copy of `messages.ReapplyBgAfterResets`

`view.go:56-62` is byte-for-byte the body of `messages/render.go:274`, which has
13 other call sites. The PR body's "entire import set is bubbletea, bubbles/key,
lipgloss, reflow/truncate and strings — no slk" is true *because* of the copy.

Your own target layout already has the answer: `internal/bubbles/ansi` is in the
tree diagram in #236. Create it in this PR and import it. If it waits, migration
#2 copies the helper again and we've reproduced F7 cause #1 inside the new tree
within two PRs — which is exactly what I flagged on the RFC about the substrate
landing ahead of the widgets.

### 3. `internal/bubbles/` is outside `boundary_test.go`'s reach

`walkGoFiles(t, ".")` runs from `internal/ui`, so a sibling `internal/bubbles`
tree is never walked. Demonstrated — I added `_ "net/http"` to
`internal/bubbles/confirmprompt/keymap.go` and
`TestTUIReachesTheAppOnlyThroughCore` still passed.

That invariant is the one AGENTS.md calls deliberate and says not to breach.
This PR doesn't breach it, but it starts quietly shrinking what it covers, and
after 29 migrations it would cover the app layer only. One-line fix: walk
`../bubbles` as well.

---

### Should also fix

**`AGENTS.md` isn't updated.** New top-level tree not in the architecture map,
which still says `internal/ui/<widget>/`; the "Known duplication" entry still
says 11 `renderBox` across 13 modal packages when it is now 10 + 1 elsewhere.
Same-commit rule, and it matters more than usual during a long migration.

**`mode_confirm_test.go` now cites code that doesn't exist** —
`mode_confirm.go:19-20`, `mode_confirm.go:21` and `confirmprompt/model.go:83`,
plus "the Code switch strips the modifier" when it's now `confirms()` doing it.
You correctly kept the row per its own `WHEN THAT DEAD CODE GOES` instruction;
it just needs re-wording. gammons/slk#188 item 1 can be closed at the same time.

### Minor

- **No `t.Parallel()` in the new package.** Parallel component tests are one of
  the headline benefits in #236 and the template PR is where it should show up.
- **`confirmPromptBox.BoxSize` dropped the explicit `(0, 0)`-when-hidden guard.**
  `lipgloss.Height("")` is 1, not 0. Unreachable today and both values dismiss
  identically, so no behaviour change — but restore it for parity with the other
  eight modals.
- **A zero-value `Model` degrades silently**: empty `KeyMap` means nothing
  matches `Confirm`, so every key cancels. `app.go:754` does call `New()`, so
  this is fine today. Given your own argument that options can't express a
  requirement, worth a doc line saying `New` is the only usable constructor.
- **Rebase** — merge-base is `4d4365b`, `main` is at `b733cce`.

### Liked

Moving `messages.BgANSI()+FgANSI()` out to `confirm.go` instead of into the
widget is exactly right, and it's the part that generalises: the host owns the
theme, the widget owns the styles. Tests went 9 → 18 plus 6 benchmarks, and the
two deliberate behaviour differences were both found, justified and written down
before I asked.

gammons:
Reviewing this as the pilot it says it is — I've kept the structural questions in #236 where they belong and stuck to what this branch does. Re-checked at `5e87b46` after your rebase. `go build`, `go vet`, `gofmt`, `golangci-lint` and `go test ./internal/bubbles/... ./internal/ui/ -race` are all green, and the 8 App-level goldens really are unchanged.

The component shape is the part I'd keep. `Styles`/`KeyMap` passed in rather than read from the theme global, `Update` returning a concrete `Model`, footer help derived from the keymap, and an import set with no slk in it — that's the right target, and the 3ms package test versus the 30s `internal/ui` suite is the argument for the whole RFC in one number.

**One thing needs fixing before this is a fair pilot: it changes behaviour, and in the direction that costs the most.**

The description says "everything else is byte-identical". `ctrl+y` and `alt+y` now **confirm** where they previously **cancelled**. Run end-to-end through the real reducer chain, same test both sides:

```
on main:      ctrl+y -> fired = false    alt+y -> fired = false
on 5e87b46:   ctrl+y -> fired = true     alt+y -> fired = true
```

The cause is the guard in `confirms()` (`internal/bubbles/confirmprompt/update.go:27-36`):

```go
if press.Text != "" || press.Mod == 0 {
    return false
}
press.Mod = 0
return key.Matches(press, m.KeyMap.Confirm)
```

`press.Text != ""` asks whether *this event* produced text, not whether the *base key* is printable. `ctrl+y` produces no text, so it passes the guard, gets its modifier stripped, and matches the `y` binding. The stated intent — "ignoring modifiers on non-printable keys" — is exactly right; `Text` just isn't the way to ask. Keying off the `Code` being a non-printable special key would be.

The two prompts this guards are **"Delete message?"** and **"Quit slk?"** (`internal/ui/app.go:3740` and `:1758`). `ctrl+y` is yank in readline-style keymaps, so this is a plausible reflex landing on a destructive action, and `openQuitConfirm`'s own comment says the prompt exists so "an accidental Ctrl+C in any mode never silently kills the app" — the prompt is a safety net, which makes a widened confirm surface the worst place to drift.

**The test named for this case provides no coverage of it.** `TestUpdateIgnoresModifiersOnSpecialKeys` builds:

```go
tea.KeyPressMsg{Code: 'z', Text: "z", Mod: tea.ModCtrl}
```

Bubbletea never produces that — a real `ctrl+z` has `Text: ""`. With `Text` set, `String()` is `"z"`, not `"ctrl+z"`, so it fails the *first* `key.Matches` and returns at the guard without ever reaching the stripping path the test is named after. It also happens to pick the one letter where the bug can't appear: `z` isn't a Confirm key, so it cancels either way. Two independent reasons it passes on broken code.

Worth a table over `{y, n, enter, esc} × {none, shift, ctrl, alt, ctrl+alt}` asserting against main's behaviour — that's the grid that would have caught it, and given the RFC's ambition it's the kind of test every migrated component will want.

**On scope, for #236 rather than here:** this moves one of the 13 modal packages and one of the 11 `renderBox` copies out of `internal/ui/`, both of which AGENTS.md cites (lines 42 and 153) and the Phase 4 shared-chrome plan is counted against. I don't think that's a problem with this PR — it's the first real data point for the RFC, and the fact that it surfaces the question is useful. AGENTS.md should be updated in whichever PR actually establishes `internal/bubbles/` as the convention, not in a pilot that may yet be reshaped. I'll take the substantive version of that discussion to #236.

No label while it's a draft. Fix the modifier guard and give it the real grid, and the component shape itself I'm happy with.

